PR #1, opened 2020-04-08
fix multiverse sources for 1604

robbi5:


em0lar:
Wouldn't it be better to use the host fact `ansible_distribution_release` instead of a specific version at every point where a release name is specified?

robbi5:
maybe yes, but currently BBB itself is pinned to 16.04. for the next bbb release the role has to be adjusted anyway, but we don't know if the support for 18.04 is introduced and 16.04 dropped then.

PR #2, opened 2020-04-12
Extended testing

juju4:
Testing in travis
* molecule config with docker driver but limitations as deployment makes extensive use of systemd and docker inside
* kitchen+lxd+testing with serverspec
* stages: lint, deploy (official xenial), alternate deployment/variant with more recent nodejs, mongodb or distribution (last one only if branch starts with devel)
  * xenial + nodejs 10.x
  * xenial + nodejs 12.x
  * xenial + mongodb 3.6
  * xenial + mongodb 4.2
  * bionic + mongodb 4.2: this one always failing as packages are not available
Likely that needs more functional tests but good start.

https://travis-ci.com/github/juju4/ansible-role-bigbluebutton

Misc
- lint
- locale management if inside docker
- self-signed certificate option for testing without letsencrypt
- fix/test for idempotency

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -13,8 +13,8 @@ galaxy_info:
   platforms:
     - name: Ubuntu
       versions:
-        - trusty
-        # - xenial
+        - xenial
+        # - bionic
 
   galaxy_tags: []
 
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -1,138 +1,20 @@
+---
 galaxy_info:
   role_name: bigbluebutton
   author: n0emis
-  description: BigBlueButton installation (following the documentation on http://docs.bigbluebutton.org/install/install.html)
+  description: >
+    BigBlueButton installation (following the documentation on
+    http://docs.bigbluebutton.org/install/install.html)
 
   license: MIT
 
   min_ansible_version: 2.0
 
   platforms:
-  #- name: EL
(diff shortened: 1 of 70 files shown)


tasks/main.yml
```diff
@@ -65,9 +72,57 @@
...
+  service:
+    name: mongod
+    enabled: true
+    state: started
+
 - import_tasks: letsencrypt.yml
   when: bbb_letsencrypt_enable
 
+- import_tasks: certificate-selfsigned.yml
+  when: not bbb_letsencrypt_enable
```
n0emis, line 124: Please make this step configurable, so that users can use pregenerated certificates with `bbb_ssl_cert` and `bbb_ssl_key`...

meta/main.yml
```diff
@@ -1,138 +1,20 @@
...
-  #  - etch
-  #  - jessie
-  #  - lenny
-  #  - sid
-  #  - squeeze
-  #  - stretch
-  #  - wheezy
+    - name: Ubuntu
+      versions:
+        - trusty
```
n0emis, line 16: ToDo (for me): remember #1 when merging...

tasks/main.yml
n0emis: @juju4 Could you take a look at this? Then I can merge your PR...

PR #3, opened 2020-04-16
Fix HTTP redirection, APT keys, remove dups

anadahz:
* Do not issue certificates for ansible_fqdn as it will not work for
  hosts that have no set a valid and registered FQDN.
* Remove duplicate of BigBlueButton APT key
* Set HTTPS endpoints for repositories that support it
* Add APT key fingerprints
* Fix nginx HTTP to HTTPS redirection
* Add Greenlight registration variable
* Remove unused variable bbb_api_demos_enable

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,11 +14,11 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_greenlight_enable` | enable installation of the greenlight client | `yes` |
 | `bbb_greenlight_secret` | Secret for greenlight _(required when using greenlight)_ |  | can be generated with `openssl rand -hex 64`
 | `bbb_greenlight_db_password` | Password for greenlight's database  _(required when using greenlight)_ | | can be generated with `openssl rand -hex 16`
+| `bbb_greenlight_default_registration` | Registration option open(default), invite or approval
 | `bbb_disable_recordings` | Disable options in gui to have recordings | `no` | [Recordings are running constantly in background](https://github.com/bigbluebutton/bigbluebutton/issues/9202) which is relevant as privacy relevant user data is stored
 | `bbb_api_demos_enable` | enable installation of the api demos | `no` |
 | `bbb_nodejs_version` | version of nodejs to be installed | `8.x` |
 | `bbb_system_locale` | the system locale to use | `en_US.UTF-8` |
-| `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, approval |
 | `bbb_cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][bbb_cpuschedule] |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
 
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_api_demos_enable` | enable installation of the api demos | `no` |
 | `bbb_nodejs_version` | version of nodejs to be installed | `8.x` |
 | `bbb_system_locale` | the system locale to use | `en_US.UTF-8` |
+| `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, aprroval |
+| `bbb_cam_rec` | Disable recording | `false` | Disable recording by default: false/true |
 
 ### Extra options for Greenlight
 The Web-Frontend has some extra configuration options, listed below:
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_system_locale` | the system locale to use | `en_US.UTF-8` |
 | `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, aprroval |
 | `bbb_cam_rec` | Disable recording | `false` | Disable recording by default: false/true |
-| `cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][cpuschedule] |
-| `freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][freeswitch_ipv6] |
+| `bbb_cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][bbb_cpuschedule] |
+| `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
 
 ### Extra options for Greenlight
 The Web-Frontend has some extra configuration options, listed below:
@@ -106,5 +106,5 @@ This is an example, of how to use this role. Warning: the value of the Variables
 ## License
 MIT
 
-[cpuschedule]: https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-start-with-a-setscheduler-error
-[freeswitch_ipv6]: https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-bind-to-port-8021
+[bbb_cpuschedule]: https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-start-with-a-setscheduler-error
+[bbb_freeswitch_ipv6]: https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-bind-to-port-8021
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_system_locale` | the system locale to use | `en_US.UTF-8` |
 | `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, aprroval |
 | `bbb_cam_rec` | Disable recording | `false` | Disable recording by default: false/true |
+| `cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][cpuschedule] |
 
 ### Extra options for Greenlight
 The Web-Frontend has some extra configuration options, listed below:
@@ -103,3 +104,5 @@ This is an example, of how to use this role. Warning: the value of the Variables
 
 ## License
 MIT
+
+[cpuschedule]: https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-start-with-a-setscheduler-error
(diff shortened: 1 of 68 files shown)


templates/greenlight/env.j2
```diff
@@ -231,4 +231,4 @@ DB_PASSWORD={{ bbb_greenlight_db_password }}
 #   open - For open registration
 #   invite - For invite only registration
 #   approval - For approve/decline registration
-DEFAULT_REGISTRATION=open
```
n0emis, line 234: Thanks so far! Could you mention this variable in the README, so it is documented somewhere?

anadahz: Done in https://github.com/n0emis/ansible-role-bigbluebutton/pull/3/commits/90e51b3f9b8401dd5d5c97da6f96fcb51267f384

tasks/config.yml
```diff
@@ -87,4 +95,29 @@
   template:
      src: bbb/bigbluebutton.j2
      dest: /etc/nginx/sites-available/bigbluebutton
-  notify: reload nginx
+  notify: restart nginx
```
n0emis, line 98: One last comment: It is not strictly necessary to completely _restart_ nginx here, a simple reload would be enough...

anadahz: This is the case but not with how BigBlueButton is being installed.

Without restarting nginx I wasn't able to view the BigBlueButton main page and I was instead seeing the demo page.
See also: https://github.com/n0emis/ansible-role-bigbluebutton/pull/3/commits/9d4e884a8ca417e0e7d776cccb426e2c67c5f139

@n0emis Let me know if this work for you.

n0emis: Yes, indeed. I did a installation on a fresh ubuntu 16.04 server, but just with reloading nginx and everything worked fine...

defaults/main.yml
```diff
@@ -14,4 +13,8 @@ bbb_ssl_cert: "/etc/letsencrypt/live/{{ bbb_hostname }}/fullchain.pem"
 bbb_ssl_key: "/etc/letsencrypt/live/{{ bbb_hostname }}/privkey.pem"
 
 
-bbb_apt_key: 328BD16D
+bbb_apt_key: 770C4267C5E63474D171B60937B5DD5EFAB46452
+mongodb_apt_key: 0C49F3730359A14518585931BC711F9BA15703C6
+
+# Greenlight registration mode: open (default), invite, approval
+bbb_greenlight_reg: open
```
n0emis, line 20: please also add the other variables you added:
```yaml
bbb_disable_rec: false
bbb_cpuschedule: true
bbb_freeswitch_ipv6: true
```

README.md
```diff
@@ -17,6 +17,10 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_api_demos_enable` | enable installation of the api demos | `no` |
 | `bbb_nodejs_version` | version of nodejs to be installed | `8.x` |
 | `bbb_system_locale` | the system locale to use | `en_US.UTF-8` |
+| `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, aprroval |
+| `bbb_cam_rec` | Disable recording | `false` | Disable recording by default: false/true |
```
n0emis, line 21: please also make the spelling of this variable consistent throughout the role...

tasks/config.yml
n0emis: Thanks to @robbi5 for pointing out: Restarting is just needed when changing something with the listen directives. But as the default-config already listend on the ports 80 and 443, we do not need to reload here...

README.md
```diff
@@ -17,6 +17,10 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_api_demos_enable` | enable installation of the api demos | `no` |
 | `bbb_nodejs_version` | version of nodejs to be installed | `8.x` |
 | `bbb_system_locale` | the system locale to use | `en_US.UTF-8` |
+| `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, aprroval |
```
anarcat, line 20:
```suggestion
| `bbb_greenlight_reg` | Registration mode | `open` | Greenlight registration mode: open, invite, approval |
```

README.md
anarcat, line 21: I would also suggest to avoid the double-negative form:

```suggestion
| `bbb_enable_rec` | Enable recording | `true` | Enable recording by default: false/true |
```

tasks/config.yml
```diff
@@ -52,6 +52,14 @@
     line: "playback_protocol: https"
   notify: restart bigbluebutton
 
+- name: disable recordings by default
+  lineinfile:
+    path: /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
+    regexp: '^disableRecordingDefault'
+    line: 'disableRecordingDefault=true'
+  notify: restart bigbluebutton
+  when: bbb_disable_rec
```
anarcat, line 61:
```suggestion
  when: not bbb_enable_rec
```

defaults/main.yml
anarcat, line 20:
```suggestion
bbb_enable_rec: true
bbb_cpuschedule: true
bbb_freeswitch_ipv6: true
```

anadahz:
Thanks for the suggestions @n0emis and @anarcat I resolved the merge conflicts and shaped the patch based on the suggestions.

tasks/config.yml
```diff
@@ -89,6 +89,32 @@
...
 
+- name: disable CPUSchedulingPolicy in FreeSWITCH service
+  lineinfile:
+    path: /lib/systemd/system/freeswitch.service
+    regexp: '^CPUSchedulingPolicy=rr'
+    line: '#CPUSchedulingPolicy=rr'
+  notify: restart freeswitch
+  when: not bbb_cpuschedule
+
+- name: disable IPv6 in FreeSWITCH service 1/2
```
robbi5, line 100: this should also only happen `when: not bbb_freeswitch_ipv6`?

anadahz: @robbi5 yes, otherwise it should not disable IPv6 in FreeSWITCH service.

Do I miss anything?

anadahz: @robbi5 Thanks, I missed this.

anadahz:
> it looks like recording controls (90e51b3) went away but i guess that's okay for a first step.
@anarcat master is moving fast :)
It was introduced in https://github.com/n0emis/ansible-role-bigbluebutton/commit/046e9fc043bfb97b6f403651b33b9eb4ebd5878c and I removed it from this PR.

templates/bbb/bigbluebutton.j2
```diff
@@ -2,9 +2,12 @@ server {
         listen 80;
         listen [::]:80;
         server_name {{ bbb_hostname }};
-        return 301 https://$host$request_uri;
-}
 
+        # If a user tries to come through http, redirect them through https
```
genofire, line 6: this server is listing on 80 without ssl ...
So your patch for this file is useless (sorry)

anadahz: @genofire Without this patch there is no redirect to HTTPS or am I missing anything?

genofire: there is the line: ` return 301 https://$host$request_uri;` which you try to remove, should solve the redirect allready

anadahz:
> there is the line: return 301 https://$host$request_uri; which you try to remove, should solve the redirect allready

@genofire This doesn't redirect properly.

See also:
https://github.com/bigbluebutton/bbb-install/issues/188

genofire:
Please take a closer look at:

https://github.com/bigbluebutton/bbb-install/blob/9f685bce3bf94a7322bde33fb207bffcf0d1132f/bbb-install.sh#L795-L801

vs. 

https://github.com/n0emis/ansible-role-bigbluebutton/blob/046e9fc043bfb97b6f403651b33b9eb4ebd5878c/templates/bbb/bigbluebutton.j2#L1-L12

anadahz:
> Please take a closer look at:
> 
> https://github.com/bigbluebutton/bbb-install/blob/9f685bce3bf94a7322bde33fb207bffcf0d1132f/bbb-install.sh#L795-L801
> 
> vs.
> 
> https://github.com/n0emis/ansible-role-bigbluebutton/blob/046e9fc043bfb97b6f403651b33b9eb4ebd5878c/templates/bbb/bigbluebutton.j2#L1-L12

@genofire Thanks, I missed that one. 
Reverted in https://github.com/n0emis/ansible-role-bigbluebutton/pull/3/commits/9dc90cccd9300ef31c3162354e353e1b6731a37b

tasks/config.yml
genofire: Is there a option to enable ipv6 again?

templates/greenlight/env.j2
```diff
@@ -231,4 +231,4 @@ DB_PASSWORD={{ bbb_greenlight_db_password }}
...
+DEFAULT_REGISTRATION={{ bbb_greenlight_reg | default('open') }}
```
genofire, line 234: Maybe the default value should be set central in `defaults/main.yml`

handlers/main.yml
```diff
@@ -4,10 +4,22 @@
...
+
 - name: restart bigbluebutton
   command: bbb-conf --restart
   become: true
 
+- name: restart freeswitch
+  systemd:
+    name: freeswitch
+    state: restarted
+    daemon_reload: yes
```
genofire, line 20: Does it need to reload systemd ?

templates/greenlight/env.j2
n0emis: Indeed, although this has already been added in #11 (with the default in the jinja2-template. I would still prefer this to be in the `defaults/main.yml`

n0emis: Ah, now i see: It is already in the defaults.yml so there is no further need to specify a default here...

tasks/config.yml
n0emis: > Is there a option to enable ipv6 again?

I would quite appreciate this option...

handlers/main.yml
anadahz: It seems so according to BBB documentation: 
https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-bind-to-port-8021

tasks/config.yml
anadahz: I have now included support to enable IPv6 in FreeSWITCH service.

handlers/main.yml
n0emis: Jes, indeed. This is needed, because `/lib/systemd/system/freeswitch.service` gets edited. But then only as a handler for `disable CPUSchedulingPolicy in FreeSWITCH service`

tasks/config.yml
```diff
@@ -89,6 +89,56 @@
      dest: /etc/nginx/sites-available/bigbluebutton
   notify: reload nginx
 
+- name: disable CPUSchedulingPolicy in FreeSWITCH service
+  lineinfile:
+    path: /lib/systemd/system/freeswitch.service
+    regexp: '^CPUSchedulingPolicy=rr'
+    line: '#CPUSchedulingPolicy=rr'
+  notify: restart freeswitch
+  when: not bbb_cpuschedule
```
n0emis, line 98: Please also make this option revertable...

handlers/main.yml
anadahz: This has been now fixed in: https://github.com/n0emis/ansible-role-bigbluebutton/pull/3/commits/669fe9562e847ccfd9f05173a74465a458fb46a2

tasks/config.yml
anadahz: This has been now fixed in: https://github.com/n0emis/ansible-role-bigbluebutton/pull/3/commits/669fe9562e847ccfd9f05173a74465a458fb46a2

PR #8, opened 2020-04-28
Add option to disable greenlight in application authentication

EdJoPaTo:
In order to enable an authentication provider and disable the in application authentication at the same time.

use `'false'` with quotes as it gets interpreted as a boolean resulting in true -> True and false -> False.
Greenlight seems to compare `ALLOW_GREENLIGHT_ACCOUNTS==true`.
Using false or False is always false but True is not equal true here and results in disabling it too which isnt intuitive. Hopefully prevent that by having quotes in the example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,13 @@ bbb_greenlight_office365:
     hd:
 ```
 
+#### In Application Authentication
+By default, the ability for anyone to create a Greenlight account is enabled. To disable this, use `false`.
+For more information see: https://docs.bigbluebutton.org/greenlight/gl-config.html#in-application-greenlight
+```yaml
+bbb_greenlight_accounts: 'false'
+```
+
 #### RECAPTCHA
 To enable reCaptcha on the user sign up, define these 2 keys.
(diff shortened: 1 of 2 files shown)


EdJoPaTo:
Thinking about it im not sure this is useful in a basic role.

As you probably require an admin account you will need the Greenlight account login at first. Then you are able to give the admin role to an account logged in from another authentication provider. Only then you can safely disable the Greenlight accounts.

Doing so might be better of by another task seperate from this basic setup.
Other thoughts on this?

n0emis:
> Thinking about it im not sure this is useful in a basic role.
> 
> As you probably require an admin account you will need the Greenlight account login at first. Then you are able to give the admin role to an account logged in from another authentication provider. Only then you can safely disable the Greenlight accounts.
> 
> Doing so might be better of by another task seperate from this basic setup.
> Other thoughts on this?

No, I think this is a good addition. In that case you could just deploy it with `ALLOW_GREENLIGHT_ACCOUNTS=true` and then later disable it, after you created an account.

PR #9, opened 2020-04-29
Option to disable recordings

EdJoPaTo:
As its [quite common](https://github.com/stadtulm/a13-ansible/blob/253d669e412e171397eb37a5695b48ff8e1cac5e/roles/bbb-ulmlernt/tasks/main.yml#L2-L21) to disable the recordings on a BBB installation I think its helpful to have this in a basic role.

This also simplifies configuration when the config changes on future versions of BBB.

I thought about having a toggle (allow recordings / disable them completely) but `autoStartRecording` only has a clear state when disabled.

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -88,3 +88,27 @@
      src: bbb/bigbluebutton.j2
      dest: /etc/nginx/sites-available/bigbluebutton
   notify: reload nginx
+
+- name: ensure default recording is disabled
+  lineinfile:
+    path: /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
+    regexp: '^disableRecordingDefault'
+    line: 'disableRecordingDefault=true'
+  when: bbb_disable_recordings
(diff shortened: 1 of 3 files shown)


tasks/config.yml
```diff
@@ -88,3 +88,27 @@
      src: bbb/bigbluebutton.j2
      dest: /etc/nginx/sites-available/bigbluebutton
   notify: reload nginx
+
+- name: ensure default recording is disabled
+  lineinfile:
+    path: /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
+    regexp: '^disableRecordingDefault'
+    line: 'disableRecordingDefault=true'
+  when: bbb_disable_recordings
```
n0emis, line 97: Thanks. Could you maybe change this, so it is possible to re-enable these options by setting the variable to true?

n0emis: I would say: when `bbb_disable_recordings` is false, all these options should be set to their default values.

PR #13, opened 2020-05-06
Improve greenlight nginx to be equal to bbb-install.sh

genofire:
`location=/ ` is duplicated.
It is installed by BBB: `/etc/bigbluebutton/nginx/greenlight-redirect.nginx`

----
Copy from new `bbb-install.sh` (syntax should be holded)

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -87,13 +87,14 @@
   file:
     path: /etc/nginx/ssl
     state: directory
-
+  when: bbb_nginx_dh
 
 - name: Generate DH Parameters with a different size (4096 bits)
   openssl_dhparam:
     path: /etc/nginx/ssl/dhp-4096.pem
     size: 4096
   notify: reload nginx
+  when: bbb_nginx_dh
 
 - name: copy bbb's nginx vhost
   template:
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -83,6 +83,18 @@
     dest: /usr/share/bbb-web/WEB-INF/classes/spring/turn-stun-servers.xml
   notify: restart bigbluebutton
 
+- name: create ssl folder for nginx
+  file:
+    path: /etc/nginx/ssl
+    state: directory
+
+
+- name: Generate DH Parameters with a different size (4096 bits)
+  openssl_dhparam:
+    path: /etc/nginx/ssl/dhp-4096.pem
+    size: 4096
+  notify: reload nginx
+
 - name: copy bbb's nginx vhost
   template:
      src: bbb/bigbluebutton.j2
--- a/templates/bbb/bigbluebutton.j2
+++ b/templates/bbb/bigbluebutton.j2
@@ -19,7 +19,12 @@ server {
     ssl_prefer_server_ciphers on;
     ssl_dhparam /etc/nginx/ssl/dhp-4096.pem;
 
+{% if bbb_nginx_privacy %}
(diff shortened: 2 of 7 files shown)


tasks/config.yml
```diff
@@ -83,6 +83,18 @@
     dest: /usr/share/bbb-web/WEB-INF/classes/spring/turn-stun-servers.xml
   notify: restart bigbluebutton
 
+- name: create ssl folder for nginx
+  file:
+    path: /etc/nginx/ssl
+    state: directory
+
+
+- name: Generate DH Parameters with a different size (4096 bits)
```
n0emis, line 92: Could you make this step optional but default enabled (since it takes quite a while) and check, if the DH-params exist before creating them?

templates/bbb/bigbluebutton.j2
```diff
@@ -1,84 +1,78 @@
...
 
+  listen 443 ssl http2;
+  listen [::]:443 ssl http2;
+  server_name {{ bbb_hostname }};
     ssl_certificate {{ bbb_ssl_cert }};
     ssl_certificate_key {{ bbb_ssl_key }};
     ssl_session_cache shared:SSL:10m;
     ssl_session_timeout 10m;
-    ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
+    ssl_protocols TLSv1.2;
```
n0emis, line 17: What about supporting TLSv1.3, is there anything that speaks against that?

templates/greenlight/greenlight-redirect.nginx.j2
```diff
@@ -0,0 +1,3 @@
+location = / {
```
n0emis, line 1: Please also make this optional (based on the `bbb_greenlight_enable` variable, as before)

tasks/config.yml
n0emis: That is also the reason, why the CI is failing. Because generating this exceeds the 10min timeout from travis...

templates/greenlight/greenlight-redirect.nginx.j2
genofire: it is optional by:
https://github.com/n0emis/ansible-role-bigbluebutton/blob/6b106944df8a09d3e8d773f2b5c8b19457205df3/tasks/main.yml#L136-L137

when this file is copied here:
https://github.com/n0emis/ansible-role-bigbluebutton/pull/13/files#diff-4a0a691a12b85e5f74cea98ea19411ceR43-R49

n0emis: :facepalm: of course...

n0emis:
Since you touch the complete nginx-vhost: could you maybe also take a look at #16 and incorporate that?

templates/bbb/bigbluebutton.j2
genofire: It would not work with v1.3 - ubuntu 16.04 LTS has OpenSSL version 1.0.2.
It required v1.1.1.
Source: https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_protocols

How do we manage it?

n0emis: Then I would say: leave it as is...

tasks/config.yml
n0emis: Thanks, looking good!

PR #16, opened 2020-05-11
Disable nginx access_log, enable error_log instead

basisbit:
### What does this PR do?
- Disabled `access_log` for nginx by setting log target `/dev/null`
- Enabled `error_log` for nginx, so that errors and critical exceptions of nginx are still logged

### Motivation
- improve data privacy by default (this is also a suggested change from https://docs.bigbluebutton.org/admin/privacy.html )
- in case the access_log is temporarily needed (for example during a DDoS), the nginx configuration can easily be changed and applied without the need to restart nginx.


basisbit:
I don't know how to ansible, sorry. Thus, I'd suggest to either merge this PR and then improve from there on, or feel free to directly commit to this branch here and thus improve the pull request.

PR #19, opened 2020-05-13
[DOC] fix coturn ports

genofire:


n0emis:
The problem here is, that nginx is already listening on port 443. I must have overlooked that when writing the role. 
I would rather change the default, because we can't have multiple services listening on one port...

genofire:
Jep, that was the problem in normal state ....

PR #33, opened 2020-05-27
use packages-eu mirror to reduce install time

basisbit:
use packages-eu mirror to reduce install time

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -29,7 +29,7 @@
 
 - name: add BigBlueButton apt key
   apt_key:
-    url: https://ubuntu.bigbluebutton.org/repo/bigbluebutton.asc
+    url: https://packages-eu.bigbluebutton.org/repo/bigbluebutton.asc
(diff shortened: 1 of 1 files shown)


genofire:
We switch already several times between the repo servers.

...

Do you like to but the URL into a variables und a default value under `defaults/main.yml`?

n0emis:
> Do you like to but the URL into a variables und a default value under `defaults/main.yml`?

I would support this, then it is easier to change for users of the role.

basisbit:
@genofire I looked at the git history and didn't see the package source being changed at all. Which commit are you referring to?

n0emis:
It was this PR: https://github.com/n0emis/ansible-role-bigbluebutton/pull/7

basisbit:
But that is not the parr which this PR was changing.
Anyways, I don't know how to ansible, and don't have the free time to learn it.
I assume this will not be merged like this and someone else will just do a PR again without attribution.

PR #39, opened 2020-06-01
configure email account verification

DO1JLR:
create option to disable e-mail account verification via e-mail.

if you set ``bbb_require_email_verification`` to ``false``, you are able to create accounts without receiving any mail.

Workaround for https://github.com/n0emis/ansible-role-bigbluebutton/issues/38

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_greenlight_secret` | Secret for greenlight _(required when using greenlight)_ |  | can be generated with `openssl rand -hex 64`
 | `bbb_greenlight_db_password` | Password for greenlight's database  _(required when using greenlight)_ | | can be generated with `openssl rand -hex 16`
 | `bbb_greenlight_default_registration` | Registration option open(default), invite or approval
-| `bbb_allow_mail_notifications`  | Require E-Mail verification for greenlight account creation | `true` |
+| `bbb_allow_mail_notifications`  | Set this to true if you want GreenLight to send verification emails upon the creation of a new account | `true` |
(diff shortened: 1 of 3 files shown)


templates/greenlight/env.j2
```diff
@@ -104,7 +104,7 @@ GOOGLE_ANALYTICS_TRACKING_ID=
 # Set this to true if you want GreenLight to send verification emails upon
 # the creation of a new account
 #
-ALLOW_MAIL_NOTIFICATIONS=true
+ALLOW_MAIL_NOTIFICATIONS={{ bbb_require_email_verification }}
```
n0emis, line 107: I would rather like this variable to be similar to the name inside the config file. I also think that the current name is a it misleading, because there is `DEFAULT_REGISTRATION` for that...

DO1JLR: good point. I will change it to ``bbb_allow_mail_notifications``

DO1JLR: I updated it and squashed my changes to one commit

README.md
```diff
@@ -24,6 +24,7 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_greenlight_secret` | Secret for greenlight _(required when using greenlight)_ |  | can be generated with `openssl rand -hex 64`
 | `bbb_greenlight_db_password` | Password for greenlight's database  _(required when using greenlight)_ | | can be generated with `openssl rand -hex 16`
 | `bbb_greenlight_default_registration` | Registration option open(default), invite or approval
+| `bbb_allow_mail_notifications`  | Require E-Mail verification for greenlight account creation | `true` |
```
n0emis, line 27: Could you maybe also change the description to be similar to the one provided in the .env?

DO1JLR: :+1: i will do that

PR #42, opened 2020-06-06
Fix molecule

0boro:


em0lar:
What are the advantages if the tests are done on a hetzner cloud vm instead of directly in a travis-vm?

0boro:
Molecule is used for local development/deployment. Is should not replace travis, because travis is used here for integration tests. Its an advantage over the Docker driver, that was used previous in Molecule by default.

BrutalBirdie:
PR Stale not ready for bbb 2.3
Closing for now.

Issue #43, opened 2020-06-08
Invalid configuration in event_socket.conf.xml


madmatah:
When applying this role multiple time, the file `/opt/freeswitch/etc/freeswitch/autoload_configs/event_socket.conf.xml` is invalid and contains multiple entries for `<param name="listen-ip" ... />`

For example : 

```
<configuration name="event_socket.conf" description="Socket Client">
  <settings>
    <param name="nat-map" value="false"/>
<param name="listen-ip" value="::"/>
    <param name="listen-port" value="8021"/>
    <param name="password" value="ClueCon"/>
    <!--<param name="apply-inbound-acl" value="loopback.auto"/>-->
    <!--<param name="stop-on-bind-error" value="true"/>-->
  </settings>
</configuration>
<param name="listen-ip" value="::"/>
<param name="listen-ip" value="::"/>
```

basisbit:
can confirm this bug. Happens on every install using this ansible role and bbb 2.2.17

basisbit:
any chance to get this fixed some time soon?

PR #44, opened 2020-06-08
Fix duplicate listen-ip entries in event_socket.conf.xml

madmatah:
As described in #43, when applying the role multiple time with the same `bbb_freeswitch_ipv6` value, duplicate lines are added to the file `event_socket.conf.xml`

This PR fixes this.

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -195,7 +195,7 @@
 - name: enable IPv6 in FreeSWITCH service 2/2
   lineinfile:
     path: /opt/freeswitch/etc/freeswitch/autoload_configs/event_socket.conf.xml
-    regexp: '<param name="listen-ip" value="127.0.0.1"/>'
+    regexp: '<param name="listen-ip" value="[a-fA-F0-9.:]+"/>'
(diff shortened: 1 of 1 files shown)


tasks/config.yml
```diff
@@ -195,7 +195,7 @@
 - name: enable IPv6 in FreeSWITCH service 2/2
   lineinfile:
     path: /opt/freeswitch/etc/freeswitch/autoload_configs/event_socket.conf.xml
-    regexp: '<param name="listen-ip" value="127.0.0.1"/>'
+    regexp: '<param name="listen-ip" value="[0-9\.:]+"/>'
```
n0emis, line 198: Could you maybe add all hexadecimal letters? then this would also math other IPv6-Addresses.

EdJoPaTo:
```suggestion
    regexp: '<param name="listen-ip" value="[a-fA-F0-9.:]+"/>'
```

and you dont need to escape the dot within the set of character (square brackets: `[]`)

madmatah: Thanks ! I updated my pull request.

PR #47, opened 2020-06-14
use nginx as a systemctl service

berott:
nginx actually is not startet after a reboot.
This fix starts nginx as a systemctl service directly after the boot


n0emis:
Looking good to me :+1:

PR #64, opened 2020-09-11
feat: make it possible to change the welcome message

toabi:
I took the BBB defaults as the defaults here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_letsencrypt_enable` | Enable letsencrypt/HTTPS | `yes` |
 | `bbb_letsencrypt_email` | E-mail for use with letsencrypt | |
 | `bbb_nginx_privacy` | only log errors not access | `yes` |
+| `bbb_default_welcome_message` | Welcome Message in the client | Welcome to <b>%%CONFNAME%%</b>!<br><br>For help on using BigBlueButton see these (short) <a href="https://www.bigbluebutton.org/html5"><u>tutorial videos</u></a>.<br><br>To join the audio bridge click the phone button.  Use a headset to avoid causing background noise for others. | Needs to be encoded with `native2ascii -encoding UTF8`! |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -35,7 +37,7 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_secret` | define the secret for bbb | `none` | `set this if you want to define the bbb-conf -secret. Otherwise the secret is generated by bbb`  
 | `bbb_cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][bbb_cpuschedule] |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
-| `bbb_freeswitch_external_ip` | Set stun server for sip and rtp on FreeSWITCH | `stun:{{ (bbb_stun_servers | first).server }}` | WARNING: the value of the default freeswitch installation is `stun:stun.freeswitch.org` |
+| `bbb_freeswitch_external_ip` | Set stun server for sip and rtp on FreeSWITCH | `stun:{{ (bbb_stun_servers \| first).server }}` | WARNING: the value of the default freeswitch installation is `stun:stun.freeswitch.org` |
```
n0emis, line 40: Please remove this slash

n0emis (changes requested):
Everything else looks fine!

README.md
EdJoPaTo: Its unrelated to the PR but without the slash the | is interpreted as table next column switch. I assume that's the reason for this change. But I assume having the slash will make the default value look strange so maybe (independently from this PR) write the default value differently.

toabi: image
 on the left with my change, on the right how it currently is…

toabi: but yeah it's not really scope of this PR so I'll remove it  so we can merge this

n0emis: Oh, sorry. I haven't noticed, that the table is broken. Well, i'll merge your PR now. But thanks for noticing...

PR #67, opened 2020-11-02
 Add bbb monitoring all in one 

e1mo:
Using `bbb_monitoring_enable` the bbb monitoring web app[^1] can be
installed on the host. It utilises the all-in-one Monitoring stack[^2] which
comes as a docker compose.

[^1]: https://github.com/greenstatic/bigbluebutton-monitoring
[^2]: https://bigbluebutton-exporter.greenstatic.dev/installation/all_in_one_monitoring_stack/

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,9 @@ Ansible role for a bigbluebutton installation (following the documentation on ht
 | `bbb_dialplan_energy_level` | Set energy level of dailplan for FreeSWITCH | `100` | only for selected profile `bbb_dialplan_quality`
 | `bbb_dialplan_comfort_noise` | Set comfort noise of dailplan for FreeSWITCH | `1400` | only for selected profile `bbb_dialplan_quality`
 | `bbb_webhooks_enable` | install bbb-webhooks | `no` |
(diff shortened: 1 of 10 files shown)


toabi:
I personally would love a quick monitoring setup, but somehow this "All In One" seems a lot… and usually when you do monitoring you have a Prometheus somewhere lurking around. 
I'd more like to easily set up an https basic-auth endpoint for the bbb-exporter to hook it up to a prometheus.

But still why not have this here too, maybe behind `bbb_monitoring_all_in_one_enbale`.

e1mo:
Hm I see your point, my intention was that it's best suited for single-server setups where one just want's to see the metrics without deploying a whole grafana-prometheus stack on their own.

I'll look into how complex it would be to add a systemd service with everything else, but I have a couple of stressful school weeks ahead so no prom-misses made.

n0emis (changes requested):
This looks good, althout (just like mentionied by @toabi ) I would also prefer to have a option of not installing this whole stack. 

I would accept this PR if you change the prefix of your variables to something like proposed by toabi. Then, if and when you have the time, I would welcome a PR just for the exporter.

n0emis (approved):

PR #68, opened 2020-11-14
Only set CPUSchedulingPolicy once

### LINKED ISSUE #59: CPUSchedulingPolicy is set in two tasks changing each other
When running two times in a row this should not change something (should be deterministic as much as possible)

Currently `CPUSchedulingPolicy` is set in two places and changing each other every time so there are changes.

https://github.com/n0emis/ansible-role-bigbluebutton/blob/54f7aa92e1c9a653d69e9e347a55760c3c65a3e9/tasks/main.yml#L33-L44

https://github.com/n0emis/ansible-role-bigbluebutton/blob/54f7aa92e1c9a653d69e9e347a55760c3c65a3e9/tasks/config.yml#L170-L188

As the second one is configurable I would think about removing the first one.
Also interesting to me, without having more knowledge into freeswitch currently: the first one does not notify freeswitch to restart as the second does so one of them is probably wrong.

e1mo:
As of af05e97 both CPUSchedulingPolicy and IOSchedulingClass were
disabled in the freeswitch systemd service for compatibility with LXD.
But however, the first option was then re-enabled in the config.yml,
triggering a restart of freeswitch every time.

But outside of virtualised environments, there is on need to remove the
CPUSchedulingPolicy option[^1]. And the IOSchedulingClass is set to
`realtime` by default[^2], but was also disabled for the LXD
compatibility.

To stop freeswitch from being restarted when there is no actual
configuration change and dial back the removal of IOSchedulingClass the
tasks where consolidated into one task so it's idempotent.

The default value for `bbb_freeswitch_ioschedule_realtime` is set to
`false` by default for backwards compatibility with setups that require
this configuration.

Close #59

[^1]: https://docs.bigbluebutton.org/2.2/troubleshooting.html#freeswitch-fails-to-start-with-a-setscheduler-error
[^2]: https://github.com/signalwire/freeswitch/blob/15ad4c23e259c1c2dff58f89e9124e5f36dd2e94/debian/freeswitch-systemd.freeswitch.service#L30


n0emis:
I am not sure, if the default of `bbb_freeswitch_ioschedule_realtime` you proposed should be the default. Since https://github.com/n0emis/ansible-role-bigbluebutton/pull/68/files#diff-f8020d83f5c78e15b9807b118f54493c15b8b217402b0c02fe644bc8a20202d4R208 deletes the line, this is a "breaking change", since freeswitch get's restarted.

Therefor I would propose to change the default of `bbb_freeswitch_ioschedule_realtime` to realtime, because this is also the default for freeswitch and is a sensible default. Although this would mean that any LXC-based deployments would break, but I think it would be sufficient to warn about this in the readme.

n0emis (approved):

PR #74, opened 2020-12-20
Allow disabling of CPUSchedulingPolicy for bbb-html5

davidmehren:
Starting in BBB v2.2.31 `bbb-html5.service` now sets `CPUSchedulingPolicy=fifo`[1].
This causes a SETSCHEDULER error when running in an LXC container.

This PR disables this option using a systemd override when `bbb_cpuschedule` is `false`. This is also the recommended solution [2].

[1] https://github.com/bigbluebutton/bigbluebutton/releases/tag/v2.2.31
[2] https://docs.bigbluebutton.org/2.2/troubleshooting.html#bbb-html5-fails-to-start-with-a-setscheduler-error

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -357,3 +357,31 @@
     replace: 'PORT=3000 /usr/share/$NODE_VERSION/bin/node main.js'
   when: not bbb_html5_node_options is defined
   notify: restart bigbluebutton
+
+- name: Create overrides directory for bbb-html5.service
+  file:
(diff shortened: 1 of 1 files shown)


n0emis:
The documentation also recommends to set `Nice=-10`. Is there a reason you did no set this here?

n0emis:
Otherwhise this PR looks good 👍🏼

davidmehren:
> The documentation also recommends to set Nice=-10. Is there a reason you did no set this here?

I think I just missed it. Fixed and rebased!

PR #76, opened 2021-01-02
add option bbb_guestpolicy - default ASK_MODERATOR

juju4:


EdJoPaTo:
The default of BBB is `defaultGuestPolicy=ALWAYS_ACCEPT` which should also be done in this role per default.

Also additional options should be documented in the readme.

n0emis (approved):

PR #79, opened 2021-01-10
Add possibility to set the default presentation

Paktosan:


EdJoPaTo:
Personally I just replaced the `default.pdf` via `copy` which results in no config change.
With this PR we can set the path but cant copy our own file over to the server.
Shouldnt this be done here also?

Personally I would expect to provide this role with a path to a file and this role does the rest.

Paktosan:
`https://docs.bigbluebutton.org/2.2/customize.html#change-the-default-presentation` describes both ways as valid.
We have chosen to implement the config change because for our scenario with multiple servers it made more sense to provide the default presentation in a central place.
I agree that a complete solution would also include the possibility to copy a file instead.
However, this was just out of scope for us.

EdJoPaTo:
I wasnt aware of it working across servers and havnt thought of that. Thanks for clarifying!

I thought about improving the readme entry then to explain that difference more easily. But I'm not sure how to do that in a useful way as the default value already contains URL so this might be more obvious. Maybe just keep it as it is.

n0emis (approved):
There is a `notify: restart bigbluebutton` missing, otherwise this looks fine. I will just add the notify after the merge...

PR #81, opened 2021-01-21
Toggle generation of mp4 recordings

e1mo:
According to the BBB Docs[^0], IOS only supports mp4 files. There is the
possibility to render recordings also to mp4, but it's disabled by
default. Using `bbb_recordings_mp4` (defaults to `false` => nothing
changed) the generation of mp4 recordings can be enabled.

[^0]: https://docs.bigbluebutton.org/2.2/customize.html#enable-playback-of-recordings-on-ios

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -72,6 +72,7 @@
   command: bbb-conf --secret
   changed_when: false
   register: result
+  check_mode: no
 
 - name: parse bbb secret
   set_fact:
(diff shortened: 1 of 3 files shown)


tasks/config.yml
```diff
@@ -72,6 +72,7 @@
   command: bbb-conf --secret
   changed_when: false
   register: result
+  check_mode: no
```
toabi, line 75: This change seems unrelated to mp4? Or what does this change/fix?

n0emis (changes requested):
could you maybe provide a way to disable webm? maybe a list of enabled video_formats?

tasks/config.yml
e1mo: Otherwise the execution in check-mode will fail. But yes, I committed that by accident and this should probably be in a separate PR.

e1mo:
I've also rebaed this PR on the current stable branch.

BrutalBirdie (changes requested):
The way this is done is not future proof imo.
See comment in code.

tasks/config.yml
```diff
@@ -357,3 +357,12 @@
...
   notify: restart bigbluebutton
+
+- name: Set recording formats
+  replace:
+    path: /usr/local/bigbluebutton/core/scripts/presentation.yml
+    regexp: '^video_formats:\s*\n\s*#?\s*-\s*webm\n\s*#*\s*-\s*mp4$'
+    replace: |-
+      video_formats:
+        {{ bbb_recordings_webm | ternary('', '#') }}- webm
+        {{ bbb_recordings_mp4 | ternary('', '#') }}- mp4
```
BrutalBirdie, line 368: IMO Using RegEx for this feels wrong.

This could be done like the Meteor setting:

https://github.com/n0emis/ansible-role-bigbluebutton/blob/56ff2f47f1568c579a963b21517c04eefb7b3ec4/tasks/config.yml#L177-L194

This would make future changes in the `/usr/local/bigbluebutton/core/scripts/presentation.yml` easier.

BrutalBirdie (commented):
something went wrong with my last review... sorry for the spam.
Here is the comment

BrutalBirdie:
PR Stale and not ready for bbb 2.3
Closing for now.
---
Awaiting Feedback.

PR #87, opened 2021-02-27
Allow modification of RTP ports

felinira:
This allows to change the RTP ports from the supplied defaults. I needed this because some of the ports on the external IP were already in use by a different service and the default port range is way too big for my needs.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -44,6 +44,11 @@ bbb_cpuschedule: true
 bbb_freeswitch_ioschedule_realtime: true
 bbb_freeswitch_ipv6: true
 bbb_freeswitch_external_ip: "stun:{{ (bbb_stun_servers | first).server }}"
+bbb_freeswitch_rtp_start_port: 16384
+bbb_freeswitch_rtp_end_port: 24576
+
+bbb_kurento_rtp_start_port: 24577
+bbb_kurento_rtp_end_port: 32768
 
 # ALWAYS_ACCEPT, ALWAYS_DENY, ASK_MODERATOR
 bbb_guestpolicy: ALWAYS_ACCEPT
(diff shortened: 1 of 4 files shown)


defaults/main.yml
```diff
@@ -44,6 +44,11 @@ bbb_cpuschedule: true
 bbb_freeswitch_ioschedule_realtime: true
 bbb_freeswitch_ipv6: true
 bbb_freeswitch_external_ip: "stun:{{ (bbb_stun_servers | first).server }}"
+bbb_freeswitch_rtp_start_port: 16384
+bbb_freeswitch_rtp_end_port: 24576
+
+bbb_kurento_rtp_start_port: 24577
+bbb_kurento_rtp_end_port: 32768
```
BrutalBirdie, line 51: Defaults looking good. 👍

BrutalBirdie (commented):
lgtm.

This is still a very specific use-case.
Still this can be helpful in this specific use-case.

Needs testing.

tasks/config.yml
```diff
@@ -284,6 +284,42 @@
...
+
+- name: change Kurento RTP end port
+  lineinfile:
+    path: /etc/kurento/modules/kurento/BaseRtpEndpoint.conf.ini
+    regexp: 'maxPort=[0-9]+'
+    line: 'maxPort={{ bbb_kurento_rtp_end_port }}'
+  notify:
+  - restart kurento
+  - reload nginx
+
```
BrutalBirdie, line 322: Upstreaming this to bbb 2.3 should not be that much work.
Also I would prefer using templates and xpath instead of lineinfile.

BrutalBirdie (changes requested):

PR #93, opened 2021-03-16
Fix FreeSWITCH IPv6 support

felinira:
It is not enough that freeswitch listens on ipv6, we must redirect incoming ipv6 packets to freeswitch via ipv6 also.
See documentation https://docs.bigbluebutton.org/2.2/troubleshooting#configure-bigbluebuttonfreeswitch-to-support-ipv6

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -193,12 +193,28 @@
     content: '{{ meteor  | to_nice_yaml }}'
     dest: /usr/share/meteor/bundle/programs/server/assets/app/config/settings.yml
 
-- name: configure freeswitch to use SSL
+- name: configure freeswitch to use SSL (IPv4 only)
   lineinfile:
     path: /etc/bigbluebutton/nginx/sip.nginx
     regexp: "proxy_pass"
     line: "        proxy_pass https://{{ ansible_default_ipv4.address }}:7443;"
   notify: reload nginx
+  when: not bbb_freeswitch_ipv6 | bool
+
+- name: configure freeswitch to use SSL (IPv4 + IPv6)
+  lineinfile:
+    path: /etc/bigbluebutton/nginx/sip.nginx
+    regexp: "proxy_pass"
+    line: "        proxy_pass https://$freeswitch_addr:7443;"
+  notify: reload nginx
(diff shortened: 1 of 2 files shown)


felinira:
All my ipv6 only traffic needed to be redirected via TURN. Now it works without TURN as well. I tested it by removing all IPv4 addresses from my computer and checking firefox about:webrtc. Before the change it needed to use TURN, after the change it didn't

felinira:
This is causing some troubles with my turn server. It doesn't work at all when this is enabled. Strange.

BrutalBirdie:
Reads good.
But since BBB & Freeswtich are still janky with IPv6 I will do some testing with this.

larsweiler:
Thanks for your Pull Request, @felinira! I did the same changes on our BBB-setup in February and after two months of testing I thought it was time to create a pull request and checked the open issues first. You were faster :)

With one user we did encounter a 1004 error and found the problem: they had an ubuntu server version installed with IPv6 privacy extensions disabled. freeswitch then received the private IPv6 address of that system, but denied that due to the `deny_private_v6` setting. The fallback with IPv4 did not work. After enabling IPv6 privacy extensions on that system, everything was fine.

Furthermore our external coturn server listens on these ports, as some firewalls blocked the default ssl port 3443:
```
listening-port=3478
tls-listening-port=443
```

These are the default ports which are even mentioned in the comments above in your `turnserver.conf.j2`, but the project's `defaults/main.yml` reads
```
bbb_coturn_port: 3478
bbb_coturn_port_tls: 3443
```

So it might be worth testing with `tls-listening-port=443`.

n0emis:
`tls-listening-port=443` might cause problems, because the nginx is already listening on that port

larsweiler:
> `tls-listening-port=443` might cause problems, because the nginx is already listening on that port

True. That's why I said, that I'm using an external turn server. On the other hand, that might not help with the freeswitch configuration changes at hand. It was just another little configuration item which might users prevent in connecting to the servers, even with IPv6 enabled.

BrutalBirdie:
PR Stale and not ready for bbb 2.3
Closing for now.
---
This could still be upstreamed to bbb 2.3

PR #95, opened 2021-03-17
Set default kurento listenting interface to ansible_default_ipv4.interface

### LINKED ISSUE #94: WebRTC seems to advertise internal docker IP addresses
Checking about:webrtc on Firefox shows me internal docker IPs like 172.18.0.1. This is a problem when combined with a turn server that rejects those IPs. It also makes the echo test take longer.

felinira:
Also adds lo because it may choose it when you have added your external address to the lo interface
Fixes #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_freeswitch_ioschedule_realtime` | [IOSchedulingClass](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingClass=) | `true` | Set this to `false` for LXD/LXC Compatibility |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
 | `bbb_freeswitch_external_ip` | Set stun server for sip and rtp on FreeSWITCH | `stun:{{ (bbb_stun_servers \| first).server }}` | WARNING: the value of the default freeswitch installation is `stun:stun.freeswitch.org` |
+| `bbb_kurento_interfaces` | Specify the listening interfaces for kurento | `{{ [ansible_default_ipv4.interface, 'lo'] }}`
 | `bbb_dialplan_quality` | Set quality of dailplan for FreeSWITCH | `cdquality` |
 | `bbb_dialplan_energy_level` | Set energy level of dailplan for FreeSWITCH | `100` | only for selected profile `bbb_dialplan_quality`
 | `bbb_dialplan_comfort_noise` | Set comfort noise of dailplan for FreeSWITCH | `1400` | only for selected profile `bbb_dialplan_quality`
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -46,6 +46,10 @@ bbb_freeswitch_ioschedule_realtime: true
 bbb_freeswitch_ipv6: true
 bbb_freeswitch_external_ip: "stun:{{ (bbb_stun_servers | first).server }}"
 
+bbb_kurento_interfaces:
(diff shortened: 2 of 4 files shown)


defaults/main.yml
```diff
@@ -46,6 +46,10 @@ bbb_freeswitch_ioschedule_realtime: true
 bbb_freeswitch_ipv6: true
 bbb_freeswitch_external_ip: "stun:{{ (bbb_stun_servers | first).server }}"
 
+bbb_kurento_interfaces:
```
n0emis, line 49: Hmm, i would like to have a default, that does not change things. Could you maybe change that?

felinira: Sure I could. I was having the problem with the default configuration that kurento was advertising the docker ip addresses on webrtc. This is a problem because the the client tries to connect to those ip addresses, fails, it takes time etc. I thought that defaulting to the default interface and lo would be a reasonable compromise, but I do see the point that it could break things. Maybe alternatively iterate over all interfaces and just filter out the docker interfaces?

n0emis: I would rather prefer the manual way (e.g. an empty default) to avoid accidentally breaking thing and rather having a good explanation (with an example) in the readme.

felinira: Ok, I understand the rationale behind that. I will do that

n0emis (approved):

tasks/config.yml
```diff
@@ -342,6 +342,24 @@
...
   - reload nginx
 
+- name: kurento only listen on specified interfaces
+  replace:
+    path: /etc/kurento/modules/kurento/WebRtcEndpoint.conf.ini
+    regexp: '^;?networkInterfaces=.*$'
+    replace: 'networkInterfaces={{ bbb_kurento_interfaces | join(",") }}'
+  when: (bbb_kurento_interfaces | default([]) | length) > 0
+  notify: restart kurento
+  tags: debug
```
BrutalBirdie, line 352: I think the `tags: debug` are still a leftover from bugging? :D
Removing them would be nice.

README.md
```diff
@@ -48,6 +48,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_freeswitch_ioschedule_realtime` | [IOSchedulingClass](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingClass=) | `true` | Set this to `false` for LXD/LXC Compatibility |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
 | `bbb_freeswitch_external_ip` | Set stun server for sip and rtp on FreeSWITCH | `stun:{{ (bbb_stun_servers \| first).server }}` | WARNING: the value of the default freeswitch installation is `stun:stun.freeswitch.org` |
+| `bbb_kurento_interfaces` | Specify the listening interfaces for kurento | `{{ [ansible_default_ipv4.interface, 'lo'] }}`
```
BrutalBirdie, line 51: Closing `|` missing.
Also the README Table consists of

Variable Name | Function | Default value | Comment
-- | -- | -- | --

 so the final block for the **Comment** `| |` is missing as well.

BrutalBirdie (changes requested):

BrutalBirdie:
@felinira sorry for asking this since you already did so much work.
Would you be able to rework this PR to fit with `bbb/2.3`?

ChrisKar96:
I am not the PR creator, but I used it successfully to overcome issue #94. Should I create a new pull request with the changes needed to be merged?

crosscodr:
> I am not the PR creator, but I used it successfully to overcome issue #94. Should I create a new pull request with the changes needed to be merged?

Hi @ChrisKar96, that would be nice, thank you!

PR #98, opened 2021-04-14
allows to disable SMTP SSL certificate valaidation

ole-db:
Allows to disable SMTP SSL certificate valaidation.
This is needed to use a local MTA on the host machine for mail delivery.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,8 @@ In that case, make sure the rest of the variables are properly set.
 
 The default value for `bbb_greenlight_smtp.sender` is `bbb@{{ bbb_hostname }}`
 
+To skip SSL certificate validation set `bbb_greenlight_smtp.openssl_verify_none` to `true`.
+
 Example Setup:
 ```yaml
 bbb_greenlight_smtp:
--- a/templates/greenlight/env.j2
+++ b/templates/greenlight/env.j2
@@ -127,6 +127,9 @@ ALLOW_MAIL_NOTIFICATIONS={{ bbb_allow_mail_notifications | ternary("true", "fals
 # Please note that enable this presents its own security risks and should not be done unless necessary.
 #   SMTP_OPENSSL_VERIFY_MODE=none
 #
+{% if bbb_greenlight_smtp.openssl_verify_none  %}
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -81,6 +81,8 @@ In that case, make sure the rest of the variables are properly set.
 
 The default value for `bbb_greenlight_smtp.sender` is `bbb@{{ bbb_hostname }}`
 
+To skip SSL certificate validation set `bbb_greenlight_smtp.openssl_verify_none` to `true`.
```
BrutalBirdie, line 84: A bit a explanation would be good when to use this.
Doc of [Action Mailer Basics](https://guides.rubyonrails.org/action_mailer_basics.html) which is used by Greenlight.
https://github.com/bigbluebutton/greenlight/blob/74cb620b1c63915b0d87456ff42d483315fec7db/config/environments/production.rb#L113-L115

The documentation states in `smtp_settings`
```
You can use the name of an OpenSSL verify constant ('none' or 'peer')
```

Warning from Greenlight .env file:

https://github.com/bigbluebutton/greenlight/blob/master/sample.env#L141-L143
```
If your mail server has a self-signed certificate, you'll also need to include the line below. 
Please note that enable this presents its own security risks and should not be done unless necessary.
```
imo:
When adding such a feature enabling both options instead of just none would be better.

templates/greenlight/env.j2
```diff
@@ -127,6 +127,9 @@ ALLOW_MAIL_NOTIFICATIONS={{ bbb_allow_mail_notifications | ternary("true", "fals
 # Please note that enable this presents its own security risks and should not be done unless necessary.
 #   SMTP_OPENSSL_VERIFY_MODE=none
 #
+{% if bbb_greenlight_smtp.openssl_verify_none  %}
+SMTP_OPENSSL_VERIFY_MODE=none
```
BrutalBirdie, line 131: If my comment on the `README.md` change makes sense, then this needs to be changed as well.
But leaving it disabled by default unless its declared is the right way imo.

BrutalBirdie (changes requested):
I would like to add these thoughts to this PR.

PR #99, opened 2021-04-20
[W.I.P] BigBlueButton 2.3 ready + features from `infra.run`

BrutalBirdie:
# WORK IN PROGRESS

I don't want this merged into `master` I would rather prefer a `master/2.3` or `develop/2.3` branch to separate between BBB 2.2 and BBB 2.3.

With this PR my goal is a working state for BigBlueButton 2.3.
I am also looking forward to merge some features from [infra.run](https://infra.run/) - [ansible-bigbluebutton-tiny](https://gitlab.com/infra.run/public/ansible-bigbluebutton-tiny/-/tree/bbb-2.3).


templates/bbb-web.properties
```diff
@@ -11,9 +11,10 @@ muteOnStart={{ bbb_mute_on_start | ternary("false", "true") }}
 appLogLevel={{ bbb_app_log_level }}
 maxInactivityTimeoutMinutes={{ bbb_meeting_inactivity_timeout_minutes }}
 defaultDialAccessNumber={{ bbb_sip_pretty_number }}
-{% if bbb_sip_pretty_number %}
-# TODO make defaultWelcomeMessageFooter customizable via variable in defaults - this way its static.
-defaultWelcomeMessageFooter=Um dieser Konferenz per Telefon beizutreten, w&auml;hlen Sie <tt>%%DIALNUM%%</tt> und geben <tt>%%CONFNUM%%</tt> als Konferenz-PIN ein.<br><br><b>Hinweis:</b> Anrufer sind ggf. stummgeschaltet. Dr&uuml;cken Sie <tt>0</tt> um zu sprechen bzw. sich stummzuschalten.'
+{% if bbb_dialin_enabled and bbb_dialin_overwrite_footer %}
+defaultWelcomeMessageFooter={{ bbb_dialin_footer }}
+{% else %}
+defaultWelcomeMessageFooter={{ bbb_default_welcome_message_footer }}
```
BrutalBirdie, line 17: I am not fluent in jinja templateing, if someone could check this? There might be a better way to do this.
Might be even completely wrong 🤷🏻‍♂️

BrutalBirdie (commented):
someone should validate my jinja templating skills :S

README.md
```diff
@@ -73,8 +73,10 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_ntp_cron_day` | Day of the month the time-sync job should run | `*`
 | `bbb_ntp_cron_hour` | Hour when the time-sync job should run | `5`
 | `bbb_ntp_cron_minute` | Minute when the time-sync job should run | `0`
-| `bbb_html5_node_options` | Allow to set extra options for node for the html5-webclient | unset | Could be used for example with https://github.com/bigbluebutton/bigbluebutton/issues/11183 ; `--max-old-space-size=4096 --max_semi_space_size=128`
+| `bbb_html5_node_options` | Allow to set extra options for node for the html5-webclient | unset | Could be used for example with <https://github.com/bigbluebutton/bigbluebutton/issues/11183> ; `--max-old-space-size=4096 --max_semi_space_size=128`
 | `bbb_meeting_inactivity_timeout_minutes` | set the default timeout in minutes | `10` |
+| `bbb_freeswitch_socket_password` | set password for freeswitch |  | Can be generated with `pwgen -s 16 1` |
```
BrutalBirdie, line 78: This variable is not default set.
Maybe add a mandatory section for these kind of variables which **must** be set by the user.

README.md
```diff
@@ -73,8 +73,10 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
...
+| `bbb_multikurento` | enable multikurento setup | `false` |
```
BrutalBirdie, line 79: Since I did not test this yet I made the default `false`.
If this feature is stable and working and is beneficial for ever instance this could be default `true`

defaults/main.yml
```diff
@@ -97,3 +97,70 @@ bbb_ntp_cron: false
...
+        public: ''
+      options:
+        failAfter: 5
+        request_timeout: 30000
+        response_timeout: 30000
+  freeswitch:
+    esl_password: "{{ bbb_freeswitch_socket_password }}"
+  log:
+    level: warn
+    filename:
```
BrutalBirdie, line 166: This section for a single / multi kurento setup definitely need a separate section in the `README.md` which explains why multi kurento might be useful und how to configure it correctly.

Need to test these default values if it works out of the box.

tasks/config.yml
```diff
@@ -12,6 +12,19 @@
...
+    src: "{{ item }}"
+    dest: "/etc/bigbluebutton/{{ item }}"
+    owner: bigbluebutton
+    group: root
+    mode: '0600'
+  loop:
+    - bbb-apps-akka.conf
+    - bbb-fsesl-akka.conf
+  notify: restart bigbluebutton
+
```
BrutalBirdie: Needed since the single / multi kurento setup feature also enables setting a `esl_password` => `{{ bbb_freeswitch_socket_password }}` for the freeswitch.

Noticed I forgot to add the `freeswitch.yml` from the tasks. next commit will add it.

tasks/multikurento.yml
```diff
@@ -0,0 +1,90 @@
...
+- name: enable kurento-media-server target
+  become: true
+  systemd:
+     name: "kurento-media-server.target"
+     enabled: true
+     state: started
+  notify:
+    - restart kurento
+
+
```
BrutalBirdie, line 90: This is from `infra.run` I did not yet test it.
NEEDS TESTING!

tasks/singlekurento.yml
```diff
@@ -0,0 +1,16 @@
...
+
+- name: deploy kurento service override configuration
+  become: true
+  template:
+    src: kurento.override
+    dest: /etc/systemd/system/kurento-media-server.service.d/override.conf
+  notify:
+    - reload systemd
+    - restart kurento
+
```
BrutalBirdie, line 16: This is from `infra.run` needs testing!

BrutalBirdie (commented):
# [W.I.P]

I did not yet test this feature!

defaults/main.yml
```diff
@@ -164,3 +164,8 @@ bbb_webrtc_sfu_multikurento:
   log:
     level: warn
     filename:
+
+# Missing var for tasks
+# bbb_sip_account:
+# bbb_sipgw:
+bbb_sip_extension: "{{ ansible_hostname }}"
```
BrutalBirdie, line 171: This needs proper documentation in the `README.md`

tasks/dialin.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+- include: dialin_standalone.yml
+  when: bbb_sipgw is not defined
+
+- include: dialin_sipgw.yml
+  when: bbb_sipgw is defined
```
BrutalBirdie, line 7: Since the var `bbb_sipgw` is not defined by default this will only get deployed when the users sets it.
This needs documentation.

templates/standalone/sip_profile.xml
```diff
@@ -0,0 +1,10 @@
+<include>
+  <gateway name="dialin">
+    <param name="proxy" value="{{ bbb_sip_account.proxy }}"/>
+    <param name="username" value="{{ bbb_sip_account.username }}"/>
+    <param name="password" value="{{ bbb_sip_account.password }}"/>
+    <param name="extension" value="{{ bbb_sip_account.extension }}"/>
+    <param name="register" value="true"/>
+    <param name="context" value="public"/>
+  </gateway>
```
BrutalBirdie, line 9: This needs proper documentation in the `README.md`

BrutalBirdie (commented):
sip and sipgw support, needs work.

tasks/bbb-web.yml
```diff
@@ -0,0 +1,36 @@
...
+    - logback.xml
+  notify:
+    - restart bbb-web
+
+- name: write custom meteor config
+  copy:
+    content: '{{ bbb_meteor | to_nice_yaml }}'
+    dest: /etc/bigbluebutton/bbb-html5.yml
+  notify:
+    - restart bbb-html5
```
BrutalBirdie, line 36: Override the Meteor settings.
There could be a conflict with:

https://github.com/BrutalBirdie/ansible-role-bigbluebutton/blob/bbb/2.3/tasks/config.yml#L87-L104

BrutalBirdie:
`default.pdf` file missing in `files/`

defaults/main.yml
```diff
@@ -85,9 +85,87 @@ bbb_dialin_mask_caller: false
 bbb_dialin_overwrite_footer: false
 bbb_dialin_footer: <br><br>To join this meeting by phone, dial:<br>  %%DIALNUM%%<br>Then enter %%CONFNUM%% as the conference PIN number.
 
+bbb_meeting_inactivity_timeout_minutes: 10
 bbb_default_presentation: ${bigbluebutton.web.serverURL}/default.pdf
 
+bbb_custom_presentation: "default.pdf"
```
n0emis, line 91: Hmm, I am not sure about this. I would rather prefer to (by default) not touch this file and only do this, when some variables is set. What is you opinion about that?

n0emis: Okay, I maybe see the point of doing this: It allows to change back to the original version of the presentation...

BrutalBirdie: To stick closer to the BBB default, yes this makes sense.

BrutalBirdie: `taks/config.yml`
```yaml
- name: set default presentation
  become: true
  copy:
    src: "{{ bbb_custom_presentation }}"
    dest: "/var/www/bigbluebutton-default/{{ bbb_custom_presentation_name | default(bbb_custom_presentation) }}"
    mode: 0644
  when: bbb_custom_presentation is defined and bbb_custom_presentation_name is defined
```
and
`defaults/main.yml`
```yaml
# bbb_custom_presentation: "default.pdf"
# set this to the name the default presentation on the host should have.
# if its not default.pdf moderators are able to delete the default presentation. Thats sometimes helpful and works fine.
# bbb_custom_presentation_name: "default.pdf"
```

BrutalBirdie: also updating the README

README.md
```diff
@@ -24,6 +24,8 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
...
 | `bbb_default_welcome_message_footer` | Footer of the welcome message | This server is running <a href="https://docs.bigbluebutton.org/" target="_blank"><u>BigBlueButton</u></a>. | Encoded as the welcome message |
 | `bbb_default_presentation` | Location of default presentation | `${bigbluebutton.web.serverURL}/default.pdf` |
+| `bbb_custom_presentation` | Overwrite the default.pdf | `None` | Location of a custom presentation will be renamed to `default.pdf` if `bbb_custom_presentation_name` is not defined - see [Ansible search paths](https://docs.ansible.com/ansible/latest/user_guide/playbook_pathing.html) for where to place your custom pdf - Example `playbooks/files/default.pdf` |
+| `bbb_custom_presentation_name` | Set a custom presentation name | `None` | Instead of overwriting the `default.pdf` setting the name will add for example the `customer.pdf` |
```
BrutalBirdie, line 28: Added doc for `bbb_custom_presentation` and `bbb_custom_presentation_name`

tasks/config.yml
```diff
@@ -275,4 +275,4 @@
     src: "{{ bbb_custom_presentation }}"
     dest: "/var/www/bigbluebutton-default/{{ bbb_custom_presentation_name | default(bbb_custom_presentation) }}"
     mode: 0644
-  when: bbb_custom_presentation | length
+  when: bbb_custom_presentation is defined
```
BrutalBirdie: Update when condition to check if the var is defined.

BrutalBirdie (commented):
@n0emis regarding the `default.pdf` override.

handlers/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+
+- name: Populate service facts
+  ansible.builtin.service_facts:
```
BrutalBirdie, line 4: Register services.

handlers/main.yml
```diff
@@ -58,6 +62,7 @@
   systemd:
     name: kurento-media-server
     state: stopped
+  when: "'kurento-media-server@' not in services"
```
BrutalBirdie, line 65: check if multi kurento services are present and then skip this step.

BrutalBirdie (commented):
Fixing a re-deploy error when re-deploying a multi-kurento setup or switching from single kurento to multi kurento.

After a short chat with `infra.run` it seems re-deploying is not really intended by them.
They always deploy a new instance.

So switching from single to multi kurento might lead to problems.
Needs testing.

tasks/thp.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+- name: delete Transparent Huge Pages service file
+  become: true
+  file:
+    path: /usr/lib/systemd/system/disable-transparent-huge-pages.service
+    state: absent
+
+- name: disable Transparent Huge Pages flag via grub
```
davidmehren, line 9: This should probably be configurable/skippable, as LXC-deployments don't have GRUB installed, which causes this task to fail

davidmehren:
@BrutalBirdie I had to add a [few fixes](https://github.com/BrutalBirdie/ansible-role-bigbluebutton/compare/bbb/2.3...davidmehren:bbb/2.3) to you branch to make it work with my LXC setup. Should I just make a PR to your branch?

I also encountered issues when deploying with `bbb_multikurento: true`.
The handler `stop kurento` fails, because `'services' is undefined`.  The handler itself also doesn't seem to make much sense: it is only called from the `add kurento-media-server@.service` task, but then should not run if a "kurento-media-server@" service exists.

davidmehren:
I just pushed a few more fixes to my branch.

The multi-kurento setup still needs some work. It's currently not possible to switch back to a single Kurento cleanly. Also the `kurento-media-server.service` gets deleted, but `bbb-conf` expects that service to exist and prints errors. It may be possible to retain that service as a dummy and just make it start the `kurento-media-server@.service` template units. I'll have to dig into the systemd docs a bit more...

BrutalBirdie:
> @BrutalBirdie I had to add a [few fixes](https://github.com/BrutalBirdie/ansible-role-bigbluebutton/compare/bbb/2.3...davidmehren:bbb/2.3) to your branch to make it work with my LXC setup. Should I just make a PR to your branch?
> 
> I also encountered issues when deploying with `bbb_multikurento: true`.
> The handler `stop kurento` fails, because `'services' is undefined`. The handler itself also doesn't seem to make much sense: it is only called from the `add kurento-media-server@.service` task, but then should not run if a "kurento-media-server@" service exists.

Thanks for that!
Looked into your changes and they make sense 👍🏼 

About the failing restarts, will look into that today.

> I just pushed a few more fixes to my branch.
>
> The multi-kurento setup still needs some work. It's currently not possible to switch back to a single Kurento cleanly. Also the > kurento-media-server.service gets deleted, but bbb-conf expects that service to exist and prints errors. It may be possible to > retain that service as a dummy and just make it start the kurento-media-server@.service template units. I'll have to dig into the > systemd docs a bit more...

Yep the multi-kurento setup is from `infra.run` and I talked (wrote?) to one of the devs.
They never intended to have a switchable state for their ansible role, also re-deploying is not really considered.

I like your idea about the `kurento-media-server.service` as a dummy service in a multi-kurento setup to just start the other kurentos.

This would make stuff a lot easier and also fix the problem with the handler failing the restart in a multi-kurento setup.

---
Thanks for helping!

BrutalBirdie (commented):
Bad practice don't do this!
This is just for testing and need to be fixed.

README.md
```diff
@@ -18,7 +18,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_state` | Install BigBlueButton to state | `present` | for updating BigBlueButton with this role use `latest` |
 | `bbb_apt_mirror` | apt repo server for BigBlueButton packages | `https://ubuntu.bigbluebutton.org` | other value would be e.g. `https://packages-eu.bigbluebutton.org` |
 | `bbb_letsencrypt_enable` | Enable letsencrypt/HTTPS | `yes` |
-| `bbb_letsencrypt_email` | E-mail for use with letsencrypt | |
+| `bbb_letsencrypt_email` | E-mail for use with letsencrypt | | _(required)_ |
```
n0emis, line 21:
```suggestion
| `bbb_letsencrypt_email` | E-mail for use with letsencrypt | | _(required when letsencrypt is enabled)_ |
```

n0emis:
TODO: move `defaultDialAccessNumber` in `dial-in/setup.yml` from `/opt/freeswitch/etc/freeswitch/dialplan/public/dialin.xml` to overrides

BrutalBirdie:
@n0emis
> TODO: move defaultDialAccessNumber in dial-in/setup.yml from /opt/freeswitch/etc/freeswitch/dialplan/public/dialin.xml to overrides

`defaultDialAccessNumber` is already done in `tasks/dial-in/setup.yml`

```yaml
- name: set defaultDialAccessNumber
  lineinfile:
    path: /usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties
    regexp: '^defaultDialAccessNumber=?'
    line: 'defaultDialAccessNumber={{ bbb_dialin_default_number }}'
  notify: restart bigbluebutton
```
But this can be deleted since `templates/bbb-web.properties` writes `defaultDialAccessNumber` via `tasks/config.yml`

```yaml
- name: Update /etc/bigbluebutton/bbb-web.properties
  become: true
  template:
    src: bbb-web.properties
    dest: /etc/bigbluebutton/bbb-web.properties
    owner: bigbluebutton
    group: root
    mode: '0600'
  notify: restart bigbluebutton
```
Template of `bbb-web.properties` does this:
```jinja
{% if bbb_sip_pretty_number is defined %}
defaultDialAccessNumber={{ bbb_sip_pretty_number }}
{% endif %}
``` 
and `bbb_sip_pretty_number` is defined nowhere :D
So I will change the template to use `bbb_dialin_default_number`

tasks/dialin.yml
```diff
@@ -0,0 +1,7 @@
+---
+
+- include: dialin_standalone.yml
```
davidmehren, line 3: The referenced `dialin_standalone.yml` and `dialin_sipgw.yml` do not exist

davidmehren (changes requested):

defaults/main.yml
```diff
@@ -85,9 +89,90 @@ bbb_dialin_mask_caller: false
...
+        public: ''
+      options:
+        failAfter: 5
+        request_timeout: 30000
+        response_timeout: 30000
+  freeswitch:
+    esl_password: "{{ bbb_freeswitch_socket_password }}"
+  log:
+    level: warn
+    filename:
```
BrutalBirdie, line 128: You mean this part right?
Yep will delete this.

tasks/dialin.yml
BrutalBirdie: Yep totally right.
Fixing right now.

templates/bbb-web.properties
n0emis: you could write a more complicated one-liner, but I think this this is a good solution

n0emis (changes requested):
please remove the unused handlers. everything else looks good 👍🏼

n0emis (approved):

PR #100, opened 2021-04-20
configure freeswitch wss IP binding

julianbarskebgw:
This PR makes sure that the freeswitch wss-binding is the same IP adress the nginx tries to connect to.

Reason for this PR:
In our environment, each server has two network interfaces (one configured with a private and the other one with a public ip address).
The nginx tried to connect to freeswitch on [public ip]:7443 (configured in line 200) while the freeswitch was listening on [private ip]:7443.
Therefore, users could not connect to the websocket and received Error 1002 in BBB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][bbb_cpuschedule], needed for LXD/LXC Compatibility |
 | `bbb_freeswitch_ioschedule_realtime` | [IOSchedulingClass](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingClass=) | `true` | Set this to `false` for LXD/LXC Compatibility |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
+| `bbb_freeswitch_ip_address` | Set IP address for FreeSWITCH's wss-binding | `{{ ansible_default_ipv4.address }}` |
 | `bbb_freeswitch_external_ip` | Set stun server for sip and rtp on FreeSWITCH | `stun:{{ (bbb_stun_servers \| first).server }}` | WARNING: the value of the default freeswitch installation is `stun:stun.freeswitch.org` |
 | `bbb_dialplan_quality` | Set quality of dailplan for FreeSWITCH | `cdquality` |
 | `bbb_dialplan_energy_level` | Set energy level of dailplan for FreeSWITCH | `100` | only for selected profile `bbb_dialplan_quality`
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][bbb_cpuschedule], needed for LXD/LXC Compatibility |
 | `bbb_freeswitch_ioschedule_realtime` | [IOSchedulingClass](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingClass=) | `true` | Set this to `false` for LXD/LXC Compatibility |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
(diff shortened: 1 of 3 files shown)


tasks/config.yml
```diff
@@ -238,6 +238,13 @@
   - restart freeswitch
   - reload systemd
 
+- name: configure freeswitch to use correct IP address for wss-binding
+  lineinfile:
+    path: /opt/freeswitch/conf/sip_profiles/external.xml
+    regexp: "<param name=\"wss-binding\"  value=\":7443\"/>"
+    line: "    <param name=\"wss-binding\"  value=\"{{ ansible_default_ipv4.address }}:7443\"/>"
```
BrutalBirdie, line 245: Setting this default to `{{ ansible_default_ipv4.address }}` might lead to future problems.
Example would be some mastermind who is only using IPv6. 😬
Another example would be users with multible public IPs.

IMO:
Replacing `{{ ansible_default_ipv4.address }}` with a default variable `{{ freeswitch_ip_address }}` in `defaults/main.yml` `freeswitch_ip_address: "{{ ansible_default_ipv4.address }}"` would be better.

This way it still defaults to `{{ ansible_default_ipv4.address }}` but there would be an option to overwrite it.
When doing this also the README needs an update and explanation when this is used.

BrutalBirdie (changes requested):
Might lead to problems in the future.

tasks/config.yml
julianbarskebgw: I didn't see using `{{ ansible_default_ipv4.address }}` as a problem first since it was already used in the nginx configuration - but you're right.

`{{ freeswitch_ip_address }}` sounds very general so I looked into where to set the IP address for freeswitch properly and found that: [](https://freeswitch.org/confluence/display/FREESWITCH/local_ip_v4)
Sadly, for some reason this breaks the ICE negotation for us in certain network situations.

So I would stay with only changing the IP address for the wss-binding but replace it with a default variable as suggested and update the README accordingly.

BrutalBirdie: Good point!
Then NGINX config must be optimized as well.

tasks/config.yml
```diff
@@ -238,11 +238,11 @@
   - restart freeswitch
   - reload systemd
 
-- name: configure freeswitch to use correct IP address for wss-binding
+- name: configure wss-binding of FreeSWITCH
```
BrutalBirdie, line 241: I would like a more verbose output, would suggest adding the `{{ bbb_freeswitch_ip_address }}` to the `- name:`.
This way without higher verbosity the user can see which IP got set.

```yaml
- name: configure IP {{ bbb_freeswitch_ip_address }} for wss-binding of FreeSWITCH
```

BrutalBirdie (changes requested):
Minor README issue and a suggestion, thanks for the update.

README.md
```diff
@@ -47,6 +47,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.2/install.html#
 | `bbb_cpuschedule` | CPUSchedulingPolicy | `true` | Disable to fix [FreeSWITCH SETSCHEDULER error][bbb_cpuschedule], needed for LXD/LXC Compatibility |
 | `bbb_freeswitch_ioschedule_realtime` | [IOSchedulingClass](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#IOSchedulingClass=) | `true` | Set this to `false` for LXD/LXC Compatibility |
 | `bbb_freeswitch_ipv6` | Enable IPv6 support in FreeSWITCH | `true` | Disable to fix [FreeSWITCH IPv6 error][bbb_freeswitch_ipv6] |
+| `bbb_freeswitch_ip_address` | Set IP address for FreeSWITCH's wss-binding | `{{ ansible_default_ipv4.address }}` |
```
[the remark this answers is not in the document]
julianbarskebgw: Still don't see that being useful but of course you never know what kind of setups others might have!
Not too sure what to write as comment, what do you think about the latest commit?

BrutalBirdie (approved):
LGTM!
Thanks for your work! ❤️

tasks/config.yml
```diff
@@ -238,6 +238,13 @@
   - restart freeswitch
   - reload systemd
 
+- name: configure IP {{ bbb_freeswitch_ip_address }} for wss-binding of FreeSWITCH
+  lineinfile:
+    path: /opt/freeswitch/conf/sip_profiles/external.xml
+    regexp: "<param name=\"wss-binding\"  value=\":7443\"/>"
```
n0emis, line 244: Please change this regex, because re-running this task could lead to the line beeing duplicated in the config.

julianbarskebgw: I was already re-running the task when testing and it never duplicated. Since the regex doesn't match anymore once `{{ 
bbb_freeswitch_ip_address }}` gets added, I don't see how that can be a problem? Even when you set  `{{ 
bbb_freeswitch_ip_address }}` to "", it will only replace the line with itself.

n0emis (approved):

PR #104, opened 2021-05-04
Feature: UFW Firewall

### LINKED ISSUE #24: firewall  ufw (and firewalld)
It would be nice to open firewall rules automatically (the known ports are at this rule).

Here are the ansible-modules:
- [ufw](https://docs.ansible.com/ansible/latest/modules/ufw_module.html)
- [firewalld](https://docs.ansible.com/ansible/latest/modules/firewalld_module.html)

Maybe we should detect, which firewall is running on the server, to use the correct ansible-module.

### LINKED ISSUE #31: reject outbound traffic to private IP address ranges
BigBlueButton by default sends data packages to all the IP addresses which the clients send to the server as local IP addresses list so the server can try to send data to there.
Hetzner seems to be annoyed by unroutable traffic. Thus, we should add a reject rule for outbound data to the following address ranges (source is the turn server default config):

- 0.0.0.0-0.255.255.255
- 10.0.0.0-10.255.255.255
- 100.64.0.0-100.127.255.255
- 127.0.0.0-127.255.255.255
- 169.254.0.0-169.254.255.255
- 127.0.0.0-127.255.255.255
- 172.16.0.0-172.31.255.255
- 192.0.0.0-192.0.0.255
- 192.0.2.0-192.0.2.255
- 192.88.99.0-192.88.99.255
- 192.168.0.0-192.168.255.255
- 198.18.0.0-198.19.255.255
- 198.51.100.0-198.51.100.255
- 203.0.113.0-203.0.113.255
- 240.0.0.0-255.255.255.255

Action reject, so that the service doesn't "wait" for a response.

BrutalBirdie:
This would also fix #31 and fix #24


BrutalBirdie:
@n0emis this PR got deployed to the internal test instance and can be checked there.

anadahz:
@BrutalBirdie thank you for working on this PR.

Have you tested if the UFW rules intefer with Docker's iptables rules?

BrutalBirdie:
> @BrutalBirdie thank you for working on this PR.
> 
> Have you tested if the UFW rules intefer with Docker's iptables rules?

I also wanted to reject 172.16.0.0/12 and 127.0.0.1/8 networks but that was problematic regarding docker (Greenlight and BBB Exporter).
Greenlight, bbb-exporter grafana etc all working with this PR.

My public testing instance can be used here: meet.bbb.ansibletest.safeserver-support.de/ (Greenlight)
and the monitoring is external with a private htpasswd for instance monitoring 
Users are:
```yaml
  - {name: 'test1', email: 'test1@server.local', password: 'changeme', type: 'user'}
  - {name: 'test2', email: 'test2@server.local', password: 'changeme', type: 'user'}
  - {name: 'test3', email: 'test3@server.local', password: 'changeme', type: 'user'}
  - {name: 'test4', email: 'test4@server.local', password: 'changeme', type: 'user'}
  ```

n0emis:
Would it be possible to have some way to remove the rules/ufw again?

BrutalBirdie:
> Would it be possible to have some way to remove the rules/ufw again?

Latest push does that.

n0emis (approved):

BrutalBirdie:
> Do I understand correctly that the coturn ports are always open, even if I do not install coturn on the system?

Yes and good point!
I will rework that so coturn ports wont be opened when disabled.

Davka (approved):

anadahz (approved):

BrutalBirdie:
Since this is kinda of a breaking change, should we disable ufw by default?

anadahz:
> Since this is kinda of a breaking change, should we disable ufw by default?

BBB will not operate correctly without ensuring having the required ports open.

It would be really neat if a test was checking if the required ports are reachable/open.

BrutalBirdie:
> > Since this is kinda of a breaking change, should we disable ufw by default?
> 
> BBB will not operate correctly without ensuring having the required ports open.
> 
> It would be really neat if a test was checking if the required ports are reachable/open.

The thought behind my comment is the following.
I myself already have a own role for ufw, other my use nftables, firewalld yada yada.
The Idea was to have ufw in this role disabled by default so no existing setup will be interrupted by the defaults.

The test for open ports would be really neat agreed.
I will do some research if I can do something like that.
But I think this can be another PR.

BrutalBirdie:
Hmm writing out loud - my thoughts again.

I want to assume that people who use this role have some knowledge about ansible and administration. 🤞🏼
Since this can not be always the case I believe security by default is something good hence enabling ufw by default is good.

PR #105, opened 2021-05-05
Implement bbb_api_demos_enable variable

e1mo:
Currently the variable `bbb_api_demos_enable` is listed in the README,
however it is not used in the playbook. Practically, the `bbb-demo`
package is not installed and the API demo is not working.

This PR adds support for the installation and, quite importantly,
removal of the `bbb-demo` apt package.

### DIFF
--- a/tasks/installation.yml
+++ b/tasks/installation.yml
@@ -27,6 +27,21 @@
   when: bbb_webhooks_enable
   notify: restart bigbluebutton
 
+- name: install bbb api demos
+  apt:
+    name: "{{ bbb_demos }}"
+    state: "{{ bbb_state }}"
+  when: bbb_api_demos_enable
+
+- name: remove bbb api demos
+  apt:
+    name: "{{ bbb_demos }}"
+    state: "absent"
+    # Using autoremove because otherwise tomcat8
+    # would stay installed and keep running
+    autoremove: true
+  when: not bbb_api_demos_enable
+
 - name: upgrade packages
   apt:
     upgrade: "{{ bbb_upgrade_packages }}"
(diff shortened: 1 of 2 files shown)


BrutalBirdie (approved):
Thanks for the PR @e1mo.
You are spot on 👍🏼

LGTM!

tasks/installation.yml
```diff
@@ -27,6 +27,21 @@
   when: bbb_webhooks_enable
   notify: restart bigbluebutton
 
+- name: install bbb api demos
+  apt:
+    name: "{{ bbb_demos }}"
+    state: "{{ bbb_state }}"
+    # Using autoremove because otherwise tomcat8
+    # would stay installed and keep running
+    autoremove: true
```
e1mo, line 36: That line is wrong here and should move down, it seems as if I was a bit tired yesterday. I will fix that during the day and would recommend not merging before I fixed that.

e1mo:
I have swapped the autoremove part. Further more, I removed the notify. From my testing a simple, and manual, `apt install bbb-demo` is enough, no reloads required. The docs don't mention a reload either: https://docs.bigbluebutton.org/2.2/install.html#5-install-api-demos-optional

BrutalBirdie:
> I have swapped the autoremove part. Further more, I removed the notify. From my testing a simple, and manual, `apt install bbb-demo` is enough, no reloads required. The docs don't mention a reload either: https://docs.bigbluebutton.org/2.2/install.html#5-install-api-demos-optional

*Hurr* 🤔 the doc from 2.2 might be outdated when using 2.3 hmm.
But since you tested it I see no problem here.

n0emis (approved):

PR #121, opened 2021-05-18
feat: encode welcome and footer message

mschwrdtnr:
Instead of manually encode the welcome  and footer message to UTF8 this task automatically do this task for you.

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -35,6 +35,12 @@
     mode: '0600'
   notify: restart bigbluebutton
 
+- name: Encode welcome message and footer message to UTF8
+  command: native2ascii -encoding UTF8 /etc/bigbluebutton/bbb-web.properties /etc/bigbluebutton/bbb-web.properties
+  become: true
+  notify:
+    - restart bigbluebutton
+
 - name: register bbb secret
   command: bbb-conf --secret
   changed_when: false


tasks/config.yml
```diff
@@ -35,6 +35,12 @@
     mode: '0600'
   notify: restart bigbluebutton
 
+- name: Encode welcome message and footer message to UTF8
+  command: native2ascii -encoding UTF8 /etc/bigbluebutton/bbb-web.properties /etc/bigbluebutton/bbb-web.properties
```
BrutalBirdie, line 39: maybe making sure that `native2ascii` is really installed would make sense.

n0emis:
I have yet to look forther into this PR, but like it is currently it will make this role unidempotent and will lead to a restart of bbb every time it is applied on the host. Please refactor this change (e.g. make this a handler or so) to avoid this...

tasks/config.yml
mschwrdtnr: Do you really think this is necessary? `native2ascii` comes with `openjdk-8-jdk` and in [vars/main.yml](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/bbb/2.3/vars/main.yml#L12) we make sure that this package is installed.

mschwrdtnr:
> 
> 
> I have yet to look further into this PR, but like it is currently it will make this role unidempotent and will lead to a restart of bbb every time it is applied on the host. Please refactor this change (e.g. make this a handler or so) to avoid this...

I don't really get what you mean and why this task should be changed to an handler. I mean this is a small task like the others in `config.yml` and should not result in restarting bbb every time?

n0emis:
> @n0emis I don't really get what you mean and why this task should be changed to an handler. I mean this is a small task like the others in `config.yml` and should not result in restarting bbb every time?

When you run a `command`-tasks, it always returns as 'changed'. therefor the handler (`restart bigbluebutton`) will be run every time this is executed.

mschwrdtnr:
> 
> When you run a `command`-tasks, it always returns as 'changed'. therefor the handler (`restart bigbluebutton`) will be run every time this is executed.

Okay got it. So what are you think will be the best solution?
- Just remove `notify`?
- Use `changed_when: false`?
- Make this a handler? But I don't really know what you mean with "make this a handler"? Should I just move this task to `handlers/main.yml`? Then I would don't know when it's right to call this handler..

tasks/config.yml
```diff
@@ -35,6 +35,12 @@
...
+  become: true
+  notify:
```
BrutalBirdie, line 41: @mschwrdtnr
The solution that comes to my mind would be the following.

Get the md5 of the file that is going to be edited.
Run the encode step.
Check the file again if changed, then restart.

Something like that could work.

n0emis: That would be one way. Although in my opinion it would be enough to create a handler which only runs when the file has changed...

n0emis:
> Should I just move this task to `handlers/main.yml`?

Well, basically yes. Please be sure to put it somewhere above `restart bigbluebutton` in that file. You can call this handler by setting `notify: handler name` at the task. Since in the case of bbb-web.properties you wan't to call multiple handlers you can write the following:
```yaml
  notify: 
    - your handler name
    - restart bigbluebutton
```

mschwrdtnr:
> 
> 
> > Should I just move this task to `handlers/main.yml`?
> 
> Well, basically yes. Please be sure to put it somewhere above `restart bigbluebutton` in that file. You can call this handler by setting `notify: handler name` at the task. Since in the case of bbb-web.properties you wan't to call multiple handlers you can write the following:
> 
> ```yaml
>   notify: 
>     - your handler name
>     - restart bigbluebutton
> ```

@n0emis I am a bit lost. I tried the way you suggested and got the problem that there is no task where I can call this handler. If I use this handler at [Update /etc/bigbluebutton/bbb-web.properties](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/bbb/2.3/tasks/config.yml#L28) the handler will always be notified. This is because this task will always be changed if we encode the file after copy on the host (the file is copied to the server and gets encoded and therefore is always different to the copied template).

I also tried to compare the checksum of the local file and the remote file. I didn't get it work because the local file is a template and not filled with the vars. So I had to make a local copy of this file and compare it.. not sure if this is a good practice..

If I am right the task [Update /etc/bigbluebutton/bbb-web.properties](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/bbb/2.3/tasks/config.yml#L28) will always return 'changed' if it has been encoded on host and therefore notify `restart bigbluebutton` will always run. Because of the "only" way to don't update the properties everytime I have to create the local `bbb-web.properties`, compare it to the remote host and run ` Update /etc/bigbluebutton/bbb-web.properties` when changed.

Am I wrong or can you comprehend my problem @n0emis @BrutalBirdie ?

n0emis:
Ah, of course. You're right. My best guess would be to try to do this encoding inside ansible. Maybe I can find some time in the next days to search for a solution...

BrutalBirdie:
closed due to inactivity.
send a message to reopen this PR.

Issue #113, opened 2021-05-11
BBB 2.3: Listen only mode (kurento) not working, ICE 1004


l3d00m:
I've set up a fresh BBB 2.3 installation on Ubuntu 18.04 with `bbb_freeswitch_ipv6: false` and `bbb_container_compat: true` based on 395585682f7b04fc3ff62ee00001b36b0276b56d

I could connect with microphone successfully and talk with other users. But connecting as "listen only" gave a ICE 1004 error. 

Running `bbb-conf --check` on that server gave the following output (*I replaced the last two ip address octets with XXX and censored the domain name*):

```
# sudo bbb-conf --check
default

BigBlueButton Server 2.3.0 (2148)
                    Kernel version: 4.15.0
                      Distribution: Ubuntu 18.04.5 LTS (64-bit)
                            Memory: 25165 MB
                         CPU cores: 8

/etc/bigbluebutton/bbb-web.properties (override for bbb-web)
/usr/share/bbb-web/WEB-INF/classes/bigbluebutton.properties (bbb-web)
       bigbluebutton.web.serverURL: https://my-domain-bbb3.de
                defaultGuestPolicy: ALWAYS_ACCEPT
                 svgImagesRequired: true

/etc/nginx/sites-available/bigbluebutton (nginx)
                       server_name: my-domain-bbb3.de
                              port: 80, [::]:8080
                              port: 443 ssl

/opt/freeswitch/etc/freeswitch/vars.xml (FreeSWITCH)
                       local_ip_v4: 178.254.XXX.XXX
                   external_rtp_ip: stun:my-domain-bbb3.de
                   external_sip_ip: stun:my-domain-bbb3.de

/opt/freeswitch/etc/freeswitch/sip_profiles/external.xml (FreeSWITCH)
                        ext-rtp-ip: $${local_ip_v4}
                        ext-sip-ip: $${local_ip_v4}
                        ws-binding: :5066
                       wss-binding: 178.254.XXX.XXX:7443

/usr/local/bigbluebutton/core/scripts/bigbluebutton.yml (record and playback)
                     playback_host: my-domain-bbb3.de
                 playback_protocol: https
                            ffmpeg: 4.2.4-1ubuntu0.1bbb2~18.04

/etc/bigbluebutton/nginx/sip.nginx (sip.nginx)
                        proxy_pass: 178.254.XXX.XXX
                          protocol: https

[...]

PR #122, opened 2021-05-18
Add sip_ip and ip for freeswitch to kurento defaults

l3d00m:
See https://github.com/ebbba-org/ansible-role-bigbluebutton/issues/113#issuecomment-843170666 for extended discussion

The reason for this PR is that sometimes the wrong IP is used inside the default kurento/webrtc config.

This override is also implemented by the bbb-install script: [sip_ip here](https://github.com/bigbluebutton/bbb-install/blob/856c5da3e62552fe218f0fb4cd23c0bd7eb65b25/bbb-install.sh#L997) and [ip here](https://github.com/bigbluebutton/bbb-install/blob/856c5da3e62552fe218f0fb4cd23c0bd7eb65b25/bbb-install.sh#L987)

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -151,6 +151,8 @@ bbb_webrtc_sfu_multikurento:
         request_timeout: 30000
         response_timeout: 30000
   freeswitch:
+    sip_ip: "{{ ansible_default_ipv4.address }}"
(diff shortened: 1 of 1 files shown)


BrutalBirdie (approved):
No objection from me.

n0emis (approved):

PR #124, opened 2021-05-19
Add separate docker task, add docker_login, update README

anadahz:
Separate all docker tasks in a different task file. This way the docker
tasks will not run if it's not required (when not bbb_greenlight_enable
or bbb_monitoring_all_in_one_enable)

- Add docker_login task to avoid rate limits
- Add variable to set docker-compose version
- Update README documentation


defaults/main.yml
```diff
@@ -220,3 +220,5 @@ bbb_firewall_ufw:
       - rule: allow
         port: 22
         proto: tcp
+
+bbb_docker_compose_version: 1.24.1
```
Davka, line 224: Would it be possible to update docker-compose to the latest release?

Davka (approved):

defaults/main.yml
anadahz: Sure, this is fixed in: https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/124/commits/8f220dee924d448b4d6f3f6d589b82e9ef6c53ed
Thank you for reviewing the PR @Davka

Davka (approved):

n0emis (approved):

PR #133, opened 2021-05-29
run ansible-lint on GitHub actions

EdJoPaTo:
There is a [GitHub action by Ansible made to run ansible-lint](https://github.com/marketplace/actions/ansible-lint) but it takes ~2min to built its docker container to run the lints. It's also not that easy to keep the version up to date (dependabot wont work for `override-deps`). Also there are some security concerns discussed in issues why tags shouldn't be used with GitHub actions which results in an harder to update action.
Using simple run commands like `pip3 install` are also simpler to understand. Because of these reasons I decided against using the ansible-lint action.

ansible-lint does not seem to feature an "error exitcode on warning" flag which might end up just warn on some things. If there are no errors just warnings the GitHub action job will succeed and warnings might be missed. Not sure how to deal with that but having a CI for running ansible-lint is already an improvement.

Currently there is no specified version range for the `pip install` command. This might end up breaking in the future.
I think this is a matter of taste. I am also fine with adding `ansible-lint<6` for example but personally I think having a failing CI is better than an outdated CI.

In my opinion this PR should only add the GitHub action and maybe some basic style / formatting things. It shouldn't be the place to fix content based on the lints which might end up changing (improving) behavior. These kind of changes should probably be discussed on its own. (Like changing tasks to be handlers instead)

You can view the output of the action [here](https://github.com/EdJoPaTo/ansible-role-bigbluebutton/actions).


BrutalBirdie (approved):
Very Nice!

Davka (approved):

PR #111, opened 2021-05-11
extend monitoring

Davka:

PR #134, opened 2021-05-30
Added bigbluebutton-exporter as Systemd service

wilkis3:
# Other changes
* Added lock against installing bigbluebutton-exporter as Docker and as Systemd service at the same time
* Changed reload monitoring handler name
# Questions
* [x] Can variable `bbb_monitoring_recordings_from_disk` rename in `bbb_monitoring_all_in_one_recordings_from_disk`?
* [x] Should variables like `bbb_monitoring_all_in_one_htpasswd` used for both environments or should they be separate?

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_monitoring_all_in_one_htpasswd_user` | The user for the htpasswd - _(required)_ if external | `Undefined` | |
 | `bbb_monitoring_all_in_one_htpasswd` | The password for the htpasswd - _(required)_ if external | `Undefined` | |
 | `bbb_monitoring_recordings_from_disk` | Collect recordings metrics by querying the disk instead of the API. See [this](https://bigbluebutton-exporter.greenstatic.dev/exporter-user-guide/#optimizations) for details. | `true` |
-| `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service(not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
+| `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service (not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
 | `bbb_monitoring_systemd_version` | Version of the (greenstatic/bigbluebutton-exporter)[https://github.com/greenstatic/bigbluebutton-exporter] | `HEAD` | Could be a branch or tag |
 | `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bigbluebutton-exporter"` | |
 | `bbb_monitoring_systemd_external` | Enable exposure to nginx | `false` | |
--- a/README.md
+++ b/README.md
@@ -74,13 +74,13 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_monitoring_all_in_one_port` | Internal Port for the monitoring werbservice | `3001` | |
 | `bbb_monitoring_all_in_one_grafana` | Enable(true)/Disable(false) the Grafana container | `true` | |
 | `bbb_monitoring_all_in_one_prometheus` | Enable(true)/Disable(false) the prometheus container | `true` | |
-| `bbb_monitoring_all_in_one_external` | Deprecated, use `bbb_monitoring_exporter_external` instead | | Can be reached under `/mon/bbb` and `/mon/node` - requires `htpasswd` and `htpasswd_user` |
-| `bbb_monitoring_all_in_one_htpasswd_user` | Deprecated, use `bbb_monitoring_exporter_htpasswd_user` instead | | |
-| `bbb_monitoring_all_in_one_htpasswd` | Deprecated, use `bbb_monitoring_exporter_htpasswd` instead | | |
+| `bbb_monitoring_all_in_one_external` | Deprecated, use `bbb_monitoring_external` instead | | Can be reached under `/mon/bbb` and `/mon/node` - requires `htpasswd` and `htpasswd_user` |
+| `bbb_monitoring_all_in_one_htpasswd_user` | Deprecated, use `bbb_monitoring_htpasswd_user` instead | | |
+| `bbb_monitoring_all_in_one_htpasswd` | Deprecated, use `bbb_monitoring_htpasswd` instead | | |
 | `bbb_monitoring_recordings_from_disk` | Collect recordings metrics by querying the disk instead of the API. See [this](https://bigbluebutton-exporter.greenstatic.dev/exporter-user-guide/#optimizations) for details. | `true` |
-| `bbb_monitoring_exporter_external` | Enable exposure to nginx | `false` | Can be reached under `/mon/bbb` and `/mon/node` - requires `htpasswd` and `htpasswd_user`. If `bbb_monitoring_systemd_enable` is enabled, no Node Exporter installation process is included |
-| `bbb_monitoring_exporter_htpasswd_user` | The user for the htpasswd - _(required)_ if external | `Undefined` | |
-| `bbb_monitoring_exporter_htpasswd` | The password for the htpasswd - _(required)_ if external | `Undefined` | |
+| `bbb_monitoring_external` | Enable exposure to nginx | `false` | Can be reached under `/mon/bbb` and `/mon/node` - requires `htpasswd` and `htpasswd_user`. If `bbb_monitoring_systemd_enable` is enabled, no Node Exporter installation process is included |
+| `bbb_monitoring_htpasswd_user` | The user for the htpasswd - _(required)_ if external | `Undefined` | |
+| `bbb_monitoring_htpasswd` | The password for the htpasswd - _(required)_ if external | `Undefined` | |
 | `bbb_monitoring_exporter_version` | Version of the BigBlueButton Exporter for docker and systemd | `latest` if docker image is enabled or `HEAD` if systemd is enabled | If `bbb_monitoring_all_in_one_enable` is enabled, the [Docker images tags](https://hub.docker.com/r/greenstatic/bigbluebutton-exporter/tags?page=1&ordering=last_updated) can be used. If `bbb_monitoring_systemd_enable` is enabled, the [Git release tags](https://github.com/greenstatic/bigbluebutton-exporter/releases) can be used. |
 | `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service (not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
 | `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bigbluebutton-exporter"` | |
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_monitoring_recordings_from_disk` | Collect recordings metrics by querying the disk instead of the API. See [this](https://bigbluebutton-exporter.greenstatic.dev/exporter-user-guide/#optimizations) for details. | `true` |
 | `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service(not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
 | `bbb_monitoring_systemd_version` | Version of the (greenstatic/bigbluebutton-exporter)[https://github.com/greenstatic/bigbluebutton-exporter] | `HEAD` | Could be a branch or tag |
-| `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bbb-exporter"` | |
+| `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bigbluebutton-exporter"` | |
 | `bbb_monitoring_systemd_external` | Enable exposure to nginx | `false` | |
 | `bbb_dialin_enabled` | enable phone dial-in, will also remove any previous dial-in configuration if set to `false`  | `false` | |
 | `bbb_dialin_provider_proxy` | IP or Domain of your SIP provider, also known as registrar | `sip.example.net` | |
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_dialplan_comfort_noise` | Set comfort noise of dailplan for FreeSWITCH | `1400` | only for selected profile `bbb_dialplan_quality` |
 | `bbb_webhooks_enable` | install bbb-webhooks | `no` | |
 | `bbb_monitoring_all_in_one_enable` | deploy [all in one monitoring stack](https://bigbluebutton-exporter.greenstatic.dev/installation/all_in_one_monitoring_stack/) (docker) | `no` |
-| `bbb_monitoring_all_in_one_version` | Version of the `greenstatic/bigbluebutton-exporter` docker image | `{{ bbb_monitoring_exporter_version }}` |  |
+| `bbb_monitoring_all_in_one_version` | Deprecated, use `bbb_monitoring_exporter_version` instead | | |
 | `bbb_monitoring_all_in_one_directory` | Directory for the docker compose files | `/root/bbb-monitoring` | |
 | `bbb_monitoring_all_in_one_port` | Internal Port for the monitoring werbservice | `3001` | |
 | `bbb_monitoring_all_in_one_grafana` | Enable(true)/Disable(false) the Grafana container | `true` | |
@@ -81,7 +81,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_monitoring_exporter_version` | Version of the BigBlueButton Exporter for docker and systemd | `latest` if docker image is enabled or `HEAD` if systemd is enabled | If `bbb_monitoring_all_in_one_enable` is enabled, the [Docker images tags](https://hub.docker.com/r/greenstatic/bigbluebutton-exporter/tags?page=1&ordering=last_updated) can be used. If `bbb_monitoring_systemd_enable` is enabled, the [Git release tags](https://github.com/greenstatic/bigbluebutton-exporter/releases) can be used. |
 | `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service (not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
 | `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bigbluebutton-exporter"` | |
-| `bbb_monitoring_systemd_external` | Enable exposure to nginx | `false` | For the external accessibility the variables `bbb_monitoring_all_in_one_htpasswd_user` and `bbb_monitoring_all_in_one_htpasswd` have to be set |
+| `bbb_monitoring_systemd_external` | Enable exposure to nginx | `false` | For the external accessibility the variables `bbb_monitoring_all_in_one_htpasswd_user` and `bbb_monitoring_all_in_one_htpasswd` have to be set. No Node Exporter installation process included |
 | `bbb_monitoring_systemd_port` | Port of bbb-exporter | `9688` | default port 9866 is defined by the exporter [itself](https://github.com/greenstatic/bigbluebutton-exporter/blob/master/bbb-exporter/settings.py#L37) |
 | `bbb_monitoring_systemd_bind_ip` | Port of bbb-exporter | `0.0.0.0` | default bind IP 0.0.0.0 is defined by the exporter [itself](https://github.com/greenstatic/bigbluebutton-exporter/blob/master/bbb-exporter/settings.py#L38) |
 | `bbb_dialin_enabled` | enable phone dial-in, will also remove any previous dial-in configuration if set to `false`  | `false` | |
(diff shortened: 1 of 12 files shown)


BrutalBirdie:
Looks OK so far.

Buy Why?
The [bbb exporter doc](https://bigbluebutton-exporter.greenstatic.dev/installation/bigbluebutton_exporter/#installation-of-bigbluebutton-exporter) states:

```
Two installation methods are supported:

    Docker (recommended)
    Systemd (not recommended)
```
Is there a valid reason why to add the systemd style version for this when the [bbb exporter doc](https://bigbluebutton-exporter.greenstatic.dev/installation/bigbluebutton_exporter/#installation-of-bigbluebutton-exporter) its not recommended?

wilkis3:
They recommend Docker because it is easier to update, reproduce builds and comes with all dependencies pro-installed. All this advantages are general Docker advantages and have not really something specially to do with the BigBlueButton Exporter.

We do not want to use Docker for the BigBlueButton Exporter.

e1mo:
The docker setup this role is currently doing is quite a lot of overhead. You do not always need the grafana that is installed with it. I've done the original docker monitoring setup in this role and did only implement docker due to time constraints. I personally would be hugely in favor of this being merged because having something not in docker would make it easier to firewall it off to be able to access it from e.G. a wireguard tunnel.

Regarding your question about the renaming:

I would vote to keep it that way since this option is also available in the stand-alone version of the exporter with the `RECORDINGS_METRICS_READ_FROM_DISK` Variable. Since this is the same feature, I would keep it at that and just use it in both cases.

BrutalBirdie:
@wilkis3 @e1mo 

> You do not always need the grafana that is installed with it. 

Yes I had the same problem and now there is the option to disable grafana / prometheus and to expose to public with htpasswd.
Also @Davka is working on another feature where you can limit which IPs are allowed to access the data.
See #111 

https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/bbb/2.3/defaults/main.yml#L81-L90

Thanks for explaining the use-cases.

README.md
```diff
@@ -77,6 +77,10 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_monitoring_all_in_one_htpasswd_user` | The user for the htpasswd - _(required)_ if external | `Undefined` | |
 | `bbb_monitoring_all_in_one_htpasswd` | The password for the htpasswd - _(required)_ if external | `Undefined` | |
 | `bbb_monitoring_recordings_from_disk` | Collect recordings metrics by querying the disk instead of the API. See [this](https://bigbluebutton-exporter.greenstatic.dev/exporter-user-guide/#optimizations) for details. | `true` |
+| `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service(not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
```
BrutalBirdie, line 80: Maybe adding a check to:
https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/bac8ed7f0c1c8f12df00b45d63d9c3cd5fe5d39b/tasks/checkvars.yml#L3
would be nice so the role fails fast when both vars are set.

BrutalBirdie (approved):
LGTM!
Would like to see this tested.
---
@wilkis3 maybe change from DRAFT to PR or you want to do more work?

README.md
e1mo: > Would like to see this tested.

I think I'll have some time later on today or tomorrow to give it a test.

Davka:
> The docker setup this role is currently doing is quite a lot of overhead. You do not always need the grafana that is installed with it. I've done the original docker monitoring setup in this role and did only implement docker due to time constraints. I personally would be hugely in favor of this being merged because having something not in docker would make it easier to firewall it off to be able to access it from e.G. a wireguard tunnel.
> 
> Regarding your question about the renaming:
> 
> I would vote to keep it that way since this option is also available in the stand-alone version of the exporter with the `RECORDINGS_METRICS_READ_FROM_DISK` Variable. Since this is the same feature, I would keep it at that and just use it in both cases.

I submitted a PR in which you can explicitly only install the BBB exporter. This can then be used to connect an external Grafana + Prometeus.

BrutalBirdie:
@e1mo and @wilkis3 can you check against @Davka PR if this would you as well?

README.md
```diff
@@ -77,6 +77,10 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
...
+| `bbb_monitoring_systemd_version` | Version of the (greenstatic/bigbluebutton-exporter)[https://github.com/greenstatic/bigbluebutton-exporter] | `HEAD` | Could be a branch or tag |
```
crosscodr, line 81: I think it does not make sense to create an additional version-variable for this, as `bbb_monitoring_all_in_one_version` exsits for the very same purpose (version of bbb-exporter).
Given the fact, that "all-in-one"-mode and "systemd"-mode mutually exclude each other, I would rather suggest to rename `bbb_monitoring_all_in_one_version` into `bbb_monitoring_bigblueblutton_exporter_version` or something like that. This could be used in both cases then.

Caveat of a shared variable is, that one would need to distinguish between "latest" and "HEAD" in the tasks.
But its simpler in the "frontend" (defaults/main.yml).

README.md
```diff
@@ -77,6 +77,10 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
...
+| `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bigbluebutton-exporter"` | |
+| `bbb_monitoring_systemd_external` | Enable exposure to nginx | `false` | |
```
crosscodr, line 83: Can you provide an explanation for this?

crosscodr (changes requested):

README.md
wilkis3: Thanks @BrutalBirdie, I added the check to checkvars.yml

wilkis3:
@BrutalBirdie @Davka 's merge request #111 looks nice and it is a good extension. It does not solve the problem to work not with Docker in general. Maybe it solves @e1mo 's problem.

Davka:
> @BrutalBirdie @Davka 's merge request #111 looks nice and it is a good extension. It does not solve the problem to work not with Docker in general. Maybe it solves @e1mo 's problem.

I still don't understand why you don't want to run that in Docker. Docker is here for LibreOffice anyway.

tasks/monitoring/systemd.yml
```diff
@@ -0,0 +1,71 @@
...
+    dest: "{{ bbb_monitoring_systemd_directory }}"
+    version: "{{ bbb_monitoring_systemd_version }}"
+
+- name: Change ownership directory
+  file:
+    path: "{{ bbb_monitoring_systemd_directory }}"
+    state: directory
+    owner: bbb-exporter
+    group: bbb-exporter
+    more: '0755'
```
BrutalBirdie, line 21: Typo, is `more` should be `mode` :D
Thanks @EdJoPaTo for the github actions ❤️

BrutalBirdie (changes requested):

tasks/monitoring/systemd.yml
EdJoPaTo: For even more (hihi) fancy tooling, there are the suggestions that can be accepted (as a bulk). Especially helpful for these kinds of small suggestions.
```suggestion
    mode: '0755'
```

Not sure why this one wants to replace a free line instead of the correct `more` line 🧐

README.md
e1mo, line 80:
```suggestion
| `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service (not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
```

README.md
e1mo, line 83: Please add `bbb_monitoring_systemd_port` and `bbb_monitoring_systemd_bind_ip` variables. They are not documented in the systemd exporter... but can be found in the code (`PORT` and `BIND_IP` environment variable):

https://github.com/greenstatic/bigbluebutton-exporter/blob/6a44d9a89028d95019231cfdaa4480b2b616e862/bbb-exporter/settings.py#L37-L38

tasks/monitoring/systemd.yml
```diff
@@ -0,0 +1,71 @@
...
+    mode: '0644'
+    remote_src: true
+
+- name: Create /etc/bigbluebutton-exporter directory
+  file:
+    path: "/etc/bigbluebutton-exporter"
+    state: directory
+    mode: '0755'
+
+- name: Copy secret file info configuration directory
```
e1mo, line 42:
```suggestion
- name: Template exporter configuration
```

tasks/monitoring/systemd.yml
```diff
@@ -0,0 +1,71 @@
...
+    dest: "/etc/bigbluebutton-exporter/settings.env"
+    mode: '0644'
+  notify: restart monitoring service
+
+- name: Copy monitoring nginx configuration file
+  template:
+    src: "monitoring/monitoring.nginx.j2"
+    dest: "/etc/bigbluebutton/nginx/monitoring.nginx"
+    mode: '0644'
+  notify: reload nginx
```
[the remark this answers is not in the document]
e1mo: And the nginx template only tests for the `bbb_monitoring_all_in_one_external` variable but not the `bbb_monitoring_systemd_external`. In my testing, I only got the `/monitoring/` location block, no matter if `bbb_monitoring_systemd_external` true or false.

And further, only the `/mon/bbb/` should be active in my opinion when `bbb_monitoring_systemd_external` is true.

e1mo (changes requested):
First of all, thanks for implementing this! I found a couple of issues, I hope it's not to nit-picky. But overall, it mostly worked (except the public nginx, but maybe I tested it wrong or misread the README?)

tasks/monitoring/systemd.yml
```diff
@@ -0,0 +1,71 @@
...
+    requirements: "{{ bbb_monitoring_systemd_directory }}/requirements.txt"
+
+- name: Copy systemd service into place
+  copy:
+    src: "{{ bbb_monitoring_systemd_directory }}/extras/systemd/bigbluebutton-exporter.service"
+    dest: /lib/systemd/system/
+    owner: root
+    group: root
+    mode: '0644'
+    remote_src: true
```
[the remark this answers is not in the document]
e1mo: Closely bound to https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/134#discussion_r643234391

e1mo:
> @e1mo and @wilkis3 can you check against @Davka PR if this would you as well?

No. Well, if I only need the exporter this is perfectly fine. But why not? 

- Having it running as a native service provides (me) more flexibility. Systemds deep integration with the system allows me to employ more fine-grained controls about the behavior of a service. All of the logs are stored in a central place making debugging easier.
- The in #111 proposed ip restrictions do not fit my need. All requests _must_ be made via https due to the http -> https redirect. But in the case of any internal VPN we are dealing with private IPs. If we would do https with those, we would send a custom http Host in the request  (because nginx only serves this content over the normal domain). 

> > @BrutalBirdie @Davka 's merge request #111 looks nice and it is a good extension. It does not solve the problem to work not with Docker in general. Maybe it solves @e1mo 's problem.
> 
> I still don't understand why you don't want to run that in Docker. Docker is here for LibreOffice anyway.

I usually like to take a step back and ask myself "is docker (or any other helper for that matter) really necessary and is the value I'm getting out of it in a proper relation to the energy invested?". In some cases (e.g. LibreOffice, Greenlight, the all in one Monitoring Stack because its entirety would be vastly out of scope) docker is a fine solution. But in other cases, I would not judge docker as being required to do the job. This is one of these cases. This PR is quite straight forward, we do not need to take a significant detour to install the service. Get the sources, the dependencies + configuration and run the service.

And not using docker makes a couple of things easier in my opinion. Coming back to my mentioned use-case (only being able to access the exporter from a wireguard interface). Firewalling a natively running process is less complex than in a docker setup. (Back reference to the nginx ip range issues)

To wrap it up, for me it comes down to have things native if its feasable due to more opportunities / less overhead. This PR is quite straight forward and should be very low in / avoid of maintenance needs and does not add a huge overhead to this role from my point of view. I would propose going on with this PR and not going to deep in the docker discussion (I tried to write this comment as neutral as I could and not fuel any "docker is{, not} useless and massive overhead" discussions. I hope this I achieved this, if not, I'm sorry. That was not my intention). I think both can co-exist happily, so can the people who implement it.

README.md
wilkis3: Yes, I added a comment to `bbb_monitoring_systemd_external`

tasks/monitoring/systemd.yml
```diff
@@ -0,0 +1,71 @@
...
+    group: root
+    mode: 0644
+  notify: reload nginx
+  when: bbb_monitoring_systemd_external
+
+- name: start monitoring
+  systemd:
+    name: bigbluebutton-exporter
+    state: started
+    enabled: true
```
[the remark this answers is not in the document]
wilkis3: @e1mo I added a deamon reload to the "Copy systemd service into place" task

Anytime when the service file changes it perform a daemon reload

tasks/monitoring/systemd.yml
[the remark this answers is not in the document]
wilkis3: Solved in [commit 9337937](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/134/commits/933793793a730528f96fa074ff2edd7388661ae1)

README.md
wilkis3: @e1mo I added the variables `bbb_monitoring_systemd_port` and `bbb_monitoring_systemd_bind_ip`
Can you please check the changes again?

tasks/monitoring/systemd.yml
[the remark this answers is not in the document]
wilkis3: @e1mo that was updated in [templates/monitoring/monitoring.nginx.j2](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/134/files#diff-57c75fd36b9b4b1768c7c0eef0e49e345458992e686a17366ff1be26e8d2b915)

README.md
wilkis3: The disadvantages outweigh and makes the Ansible code not better readable.

README.md
wilkis3: @e1mo I added the variables bbb_monitoring_systemd_port and bbb_monitoring_systemd_bind_ip

tasks/monitoring/systemd.yml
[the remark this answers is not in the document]
wilkis3: @e1mo I updated this block in [templates/monitoring/monitoring.nginx.j2](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/134/files#diff-57c75fd36b9b4b1768c7c0eef0e49e345458992e686a17366ff1be26e8d2b915)

README.md
wilkis3: @crosscodr What do you think about the following example for the default file?
```
bbb_monitoring_exporter_version: {% if bbb_monitoring_systemd_enable | bool %}HEAD{% endif %}{% if bbb_monitoring_all_in_one_enable | bool %}master{% endif %}
```

crosscodr: Good idea, but jinja can be more elegant:
`bbb_monitoring_exporter_version: {{ 'HEAD' if bbb_monitoring_systemd_enable | bool else 'latest' }}`
I think we can leave the version to the current default if bbb_monitoring_systemd_enable is false.

wilkis3: Thanks @crosscodr, that works

wilkis3:
@e1mo @BrutalBirdie is there anything else that should be changed?

BrutalBirdie (approved):

templates/monitoring/monitoring.nginx.j2
```diff
@@ -17,4 +22,4 @@ location /mon/bbb/ {
   auth_basic_user_file /etc/bigbluebutton/nginx/.htpasswd;
   include proxy_params;
 }
-{% endif %}
\ No newline at end of file
+{% endif %}
```
crosscodr, line 25: This is a bit complicated. Let's simplify this:

```
{% if bbb_monitoring_all_in_one_external %}
location /mon/node/ {
  proxy_pass http://127.0.0.1:9100/metrics;
  auth_basic "BigBlueButton Exporter";
  auth_basic_user_file /etc/bigbluebutton/nginx/.htpasswd;
  include proxy_params;
}

location /mon/bbb/ {
  proxy_pass http://127.0.0.1:9688/;
  auth_basic "BigBlueButton Exporter";
  auth_basic_user_file /etc/bigbluebutton/nginx/.htpasswd;
  include proxy_params;
}
{% else %}
{% set bbb_monitoring_exporter_port = (bbb_monitoring_all_in_one_enabled | bool
    | ternary(bbb_monitoring_all_in_one_port, bbb_monitoring_systemd_port, 9688) %}
location /monitoring/ {
  proxy_pass http://127.0.0.1:{{ bbb_monitoring_exporter_port }}/;
  include proxy_params;
}
{% endif %}
```
This is not tested, yet :D

tasks/monitoring/systemd.yml
```diff
@@ -0,0 +1,89 @@
...
+    mode: 0644
+  when: bbb_monitoring_systemd_external
+  notify:
+    - reload nginx
+
+- name: start monitoring
+  systemd:
+    name: bigbluebutton-exporter
+    state: started
+    enabled: true
```
crosscodr, line 89: In my opinion this is a bit verbose, since it duplicates many of the tasks from "all_in_one.yml". We might do a monitoring/common.yml @Davka @BrutalBirdie?

crosscodr (changes requested):

templates/monitoring/monitoring.nginx.j2
```diff
@@ -13,8 +13,7 @@ location /mon/bbb/ {
   include proxy_params;
 }
 {% else %}
-{% set bbb_monitoring_exporter_port = (bbb_monitoring_all_in_one_enabled | bool
-    | ternary(bbb_monitoring_all_in_one_port, bbb_monitoring_systemd_port, 9688)) %}
+{% set bbb_monitoring_exporter_port = (bbb_monitoring_all_in_one_enabled true | ternary(bbb_monitoring_all_in_one_port, bbb_monitoring_systemd_port, 9688)) %}
```
crosscodr, line 16: @wilkis3 Thank you for fixing my suggestion!

e1mo:
Sorry for my delay, due to school I probably will only find time for a review not before tomorrow or end of the week.

README.md
```diff
@@ -69,7 +69,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_dialplan_comfort_noise` | Set comfort noise of dailplan for FreeSWITCH | `1400` | only for selected profile `bbb_dialplan_quality` |
 | `bbb_webhooks_enable` | install bbb-webhooks | `no` | |
 | `bbb_monitoring_all_in_one_enable` | deploy [all in one monitoring stack](https://bigbluebutton-exporter.greenstatic.dev/installation/all_in_one_monitoring_stack/) (docker) | `no` |
-| `bbb_monitoring_all_in_one_version` | Version of the `greenstatic/bigbluebutton-exporter` docker image | `latest` | |
+| `bbb_monitoring_all_in_one_version` | Version of the `greenstatic/bigbluebutton-exporter` docker image | `{{ bbb_monitoring_exporter_version }}` |  |
```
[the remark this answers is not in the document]
wilkis3: @crosscodr I added a deprecated description for `bbb_monitoring_all_in_one_version`

tasks/monitoring/systemd.yml
wilkis3: @crosscodr there is now a main.yml file that handles the duplicate tasks

templates/monitoring/monitoring.nginx.j2
wilkis3: @crosscodr the 'd' of bbb_monitoring_all_in_one_enable**d** was too much.

crosscodr (changes requested):
Hi @wilkis3, thank you for your work and your adjustments!
I have two little ones left :)

README.md
```diff
@@ -78,6 +78,12 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
...
+| `bbb_monitoring_exporter_version` | Version of the BigBlueButton Exporter for docker and systemd | `latest` if docker image is enabled or `HEAD` if systemd is enabled | If `bbb_monitoring_all_in_one_enable` is enabled, the [Docker images tags](https://hub.docker.com/r/greenstatic/bigbluebutton-exporter/tags?page=1&ordering=last_updated) can be used. If `bbb_monitoring_systemd_enable` is enabled, the [Git release tags](https://github.com/greenstatic/bigbluebutton-exporter/releases) can be used. |
+| `bbb_monitoring_systemd_enable` | Deploy monitoring as systemd service (not recommended) | `false` | Works only when `bbb_monitoring_all_in_one_enable` is `false` |
+| `bbb_monitoring_systemd_directory` | Installation directory for git repository | `"/opt/bigbluebutton-exporter"` | |
+| `bbb_monitoring_systemd_external` | Enable exposure to nginx | `false` | For the external accessibility the variables `bbb_monitoring_all_in_one_htpasswd_user` and `bbb_monitoring_all_in_one_htpasswd` have to be set. No Node Exporter installation process included |
```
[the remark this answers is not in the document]
wilkis3: @crosscodr yes, I removed `bbb_monitoring_systemd_external`, `bbb_monitoring_all_in_one_external`, `bbb_monitoring_all_in_one_htpasswd_user` and `bbb_monitoring_all_in_one_htpasswd`.

And added `bbb_monitoring_external`, `bbb_monitoring_htpasswd` and `bbb_monitoring_htpasswd_user`

crosscodr:
Hi @wilkis3, can you merge bbb/2.3 once again and remove the rest of `bbb_baseurl` from tasks/main.yml?
I will merge this PR afterwards.

wilkis3:
Hi @crosscodr, I merged bbb/2.3 again. I doesn't found any existing `bbb_baseurl` things.

crosscodr:
> Hi @crosscodr, I merged bbb/2.3 again. I doesn't found any existing `bbb_baseurl` things.

Hi @wilkis3, they are in line 13 and 55 of tasks/main.yml

wilkis3:
> > Hi @crosscodr, I merged bbb/2.3 again. I doesn't found any existing `bbb_baseurl` things.
> 
> Hi @wilkis3, they are in line 13 and 55 of tasks/main.yml

`bbb_baseurl` was found and removed!

crosscodr (approved):

PR #140, opened 2021-05-31
added check for running meetings

einhirn:
After accidentally killing a live server with an ansible run (limited to the wrong pattern) I decided to add a check for running sessions to my old playbook for BBB2.2.

Now I've added that check here.

### DIFF
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -4,7 +4,8 @@
       url: "https://{{ bbb_hostname }}/bigbluebutton/api/getMeetings?checksum={{ ('getMeetings' + bbb_secret) | hash('sha1') }}"
       return_content: true
     register: bbb_getmeetings
-    ignore_errors: true
+    failed_when: false
+    changed_when: false
     tags: always
 
   - block:
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -1,0 +1,18 @@
+---
+  - name: Check for running meetings
+    uri:
+      url: "https://{{ bbb_hostname }}/bigbluebutton/api/getMeetings?checksum={{ ('getMeetings' + bbb_secret) | hash('sha1') }}"
+      return_content: true
+    register: bbb_getmeetings
+    ignore_errors: true
+    when: bbb_check_for_running_meetings
+    tags: always
+
+  - block:
+      - name: There are running sessions on this host
+        fail: msg="There are running meetings on {{ bbb_hostname }}, skipping."
+      - name: Stop play for Hosts with running meetings
+        meta: end_host
+    when: "bbb_check_for_running_meetings and not ansible_check_mode and not bbb_getmeetings.failed \
+           and '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
+    tags: always
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -15,5 +15,5 @@
     when:
       - not ansible_check_mode
       - not bbb_getmeetings.failed
-      - '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
+      - '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content
     tags: always
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -5,7 +5,6 @@
       return_content: true
     register: bbb_getmeetings
     ignore_errors: true
-    when: bbb_check_for_running_meetings
     tags: always
 
   - block:
@@ -13,6 +12,8 @@
         fail: msg="There are running meetings on {{ bbb_hostname }}, skipping."
       - name: Stop play for Hosts with running meetings
         meta: end_host
-    when: "bbb_check_for_running_meetings and not ansible_check_mode and not bbb_getmeetings.failed \
-           and '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
+    when:
+      - not ansible_check_mode
+      - not bbb_getmeetings.failed
+      - '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
     tags: always
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -15,5 +15,5 @@
     when:
       - not ansible_check_mode
       - not bbb_getmeetings.failed
-      - '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content
+      - "'<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
     tags: always
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,6 +2,8 @@
 
 - import_tasks: checkvars.yml
 
+- import_tasks: check-for-sessions.yml
+
 - import_tasks: firewall/ufw_enable.yml
   when: bbb_firewall_ufw['enabled']
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -3,6 +3,7 @@
 - import_tasks: checkvars.yml
 
 - import_tasks: check-for-sessions.yml
+  when: bbb_check_for_running_meetings
 
 - import_tasks: firewall/ufw_enable.yml
   when: bbb_firewall_ufw['enabled']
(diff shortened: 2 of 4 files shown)


e1mo (changes requested):
Overall, I am happy about not pulling BBBs away from my users, but I would propose to add a variable (default `false`) for toggling these checks.

einhirn:
Yes, this should be optional but maybe enabled by default?

Before this, I had a way to only run "restart" etc. Handlers when I had a variable set - and of course at some point I just kept the "-e bbb_doit=yes" in the command line at all times, so the accident had to happen...

Do you have a suggestion how that variable should be called? Would "bbb_check_for_sessions" fit the bill?

EdJoPaTo:
`bbb_check_for_running_meetings` sounds better to me. A `session` is not as clear to me as compared to a "running" or "ongoing" meeting.

einhirn:
Ok, I added that variable with the better name now.

Davka (approved):

tasks/check-for-sessions.yml
```diff
@@ -0,0 +1,18 @@
+---
+  - name: Check for running meetings
+    uri:
+      url: "https://{{ bbb_hostname }}/bigbluebutton/api/getMeetings?checksum={{ ('getMeetings' + bbb_secret) | hash('sha1') }}"
+      return_content: yes
+    register: bbb_getMeetings
+    delegate_to: localhost
```
n0emis, line 7: Is there a reason this is run on localhost and not on the target host? I would prefer that this task is also run on the target, for situations where the controller can't directly access the host

einhirn: I don't know why I chose to run this on the controller - I just changed that.

Another issue seems to be that "geerlingguy.docker" and "geerlingguy.nodejs" run before this check gets executed. I don't yet know why or where these dependency roles are started...

einhirn:
I found another issue: "geerlingguy.docker" and "geerlingguy.nodejs" are executed before this check, because Ansible loads all roles including dependencies defined in meta/main.yml and runs those first... I guess that would mean refactoring to "include_role", so the other roles are executed at a defined time instead of the current behaviour.

BrutalBirdie:
> I found another issue: "geerlingguy.docker" and "geerlingguy.nodejs" are executed before this check, because Ansible loads all roles including dependencies defined in meta/main.yml and runs those first... I guess that would mean refactoring to "include_role", so the other roles are executed at a defined time instead of the current behaviour.

I am pretty sure my checkvars tasks is getting executed before anything else is done, to provide a fail fast intro instead of failing mid runtime.
https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/90f41a0d39c579e92bdbefc36c78e8bc0cb13275/tasks/checkvars.yml

Maybe you can try adding your check there in a similar fashion?

einhirn:
@BrutalBirdie the way the prerequisites are included, makes them run before anything of this role. So gerlingguy.docker or gerlingguy.nodejs might change something on the target and perhaps disrupt running BBB meetings. And I've already inserted "check-for-sessions.yml" right after check_vars (which still runs after docker and nodejs roles)...

The other issue: 
The task won't fail in case you're running the playbook in "check" mode (`-C`) or if the "uri" request does not complete at all (e.g. when there's no BBB installed yet) - look here: https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/ac33d8a90c7f0c55c40477474790180b070a53f0/tasks/check-for-sessions.yml#L15

einhirn:
@BrutalBirdie the way the prerequisites are included, makes them run before anything of this role. So gerlingguy.docker or gerlingguy.nodejs might change something on the target and perhaps disrupt running BBB meetings.

The other issue:

einhirn:
According to [Stackoverflow](https://stackoverflow.com/questions/29258759/how-to-specify-ansible-pretasks-for-a-role) there are two ways to do this: Either I remove this from "tasks/main.yml" and document how people could add this to the "pre_tasks" section in their Playbooks. But at that point a lot of the needed variables aren't defined yet, so not really.
Or we'd need to move the geerlingguy-Roles to "tasks/main.yml" via "include_role" (Ansible >=2.2) or "import_role" (Ansible >=2.4). Means quite a change in the playbook and perhaps dropping support for ancient Ansible versions - can I request that?

tasks/main.yml
```diff
@@ -2,6 +2,8 @@
 
 - import_tasks: checkvars.yml
 
+- import_tasks: check-for-sessions.yml
```
BrutalBirdie, line 5: Instead of checking on every task inside `check-for-sessions.yml` for `bbb_check_for_running_meetings` why not check for that here?
Like below for the `bbb_firewall_ufw['enabled']`.

```suggestion
- import_tasks: check-for-sessions.yml
  when: bbb_check_for_running_meetings
```

tasks/check-for-sessions.yml
```diff
@@ -0,0 +1,18 @@
...
+    when: bbb_check_for_running_meetings
+    tags: always
+
+  - block:
+      - name: There are running sessions on this host
+        fail: msg="There are running meetings on {{ bbb_hostname }}, skipping."
+      - name: Stop play for Hosts with running meetings
+        meta: end_host
+    when: "bbb_check_for_running_meetings and not ansible_check_mode and not bbb_getmeetings.failed \
+           and '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
```
BrutalBirdie, line 17: Proposal:
```suggestion
    when:
      - not ansible_check_mode
      - not bbb_getmeetings.failed
      - '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
```
https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html

```
You can use logical operators to combine conditions. When you have multiple conditions that all need to be true (that is, a logical and), you can specify them as a list:
~~~
tasks:
  - name: Shut down CentOS 6 systems
    ansible.builtin.command: /sbin/shutdown -t now
    when:
      - ansible_facts['distribution'] == "CentOS"
      - ansible_facts['distribution_major_version'] == "6"
~~~
```

tasks/check-for-sessions.yml
```diff
@@ -0,0 +1,18 @@
+---
+  - name: Check for running meetings
+    uri:
+      url: "https://{{ bbb_hostname }}/bigbluebutton/api/getMeetings?checksum={{ ('getMeetings' + bbb_secret) | hash('sha1') }}"
+      return_content: true
+    register: bbb_getmeetings
+    ignore_errors: true
+    when: bbb_check_for_running_meetings
```
BrutalBirdie, line 8:
```suggestion

```

tasks/check-for-sessions.yml
```diff
@@ -0,0 +1,19 @@
+---
+  - name: Check for running meetings
+    uri:
+      url: "https://{{ bbb_hostname }}/bigbluebutton/api/getMeetings?checksum={{ ('getMeetings' + bbb_secret) | hash('sha1') }}"
+      return_content: true
+    register: bbb_getmeetings
+    ignore_errors: true
+    tags: always
```
BrutalBirdie, line 8: I am not happy with this solution.
`ignore_errors: true` is considered a bad practice for a reason.

The current output when running this PR on a new hosts creates this error:
```
TASK [ansible-role-bigbluebutton : Check for running meetings] ******************************************************************************
fatal: [meet.bbb.ansibletest.safeserver-support.de]: FAILED! => {"changed": false, "content": "", "elapsed": 0, "msg": "Status code was -1 and not [200]: Request failed: <urlopen error [Errno 111] Connection refused>", "redirected": false, "status": -1, "url": "https://meet.bbb.ansibletest.safeserver-support.de/bigbluebutton/api/getMeetings?checksum=REDACTED"}
...ignoring
```
At least this is the error I got.
I would prefer a [`failed_when:`](https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html) condition.

Your thoughts?

BrutalBirdie (changes requested):

tasks/check-for-sessions.yml
einhirn: The idea is to never have that task fail because that would stop the ansible run. I refer to the result in the next task and stop the play for a specific host, if needed. Perhaps we can come up with something like `failed_when: never`.

einhirn: I just found https://medium.com/@sbarnea/why-ansible-ignore-errors-is-evil-500fb6e81229 which answered my question "why is it bad practice" and supported my Idea of `failed_when: never` - just by way of:
```
- name: Do not count this as a failure
  ansible.builtin.command: /bin/false
  failed_when: false
  changed_when: false

BrutalBirdie (approved):

PR #146, opened 2021-06-02
only install certbot-repo when letsencrypt is used

crosscodr:


crosscodr:
Rebased this PR.

tasks/repositories.yml
```diff
@@ -30,3 +31,4 @@
     - ppa:bigbluebutton/support
     - ppa:rmescandon/yq
     - ppa:certbot/certbot
+  when: (item == 'ppa:certbot/certbot' and bbb_letsencrypt_enable | bool) or item != 'ppa:certbot/certbot'
```
BrutalBirdie, line 34: Doesn't this block the installation of:
```yaml
 - ppa:bigbluebutton/support
- ppa:rmescandon/yq
```
when `bbb_letsencrypt_enable` is `false`?

---

Thinking . . . 🤔
nah wait the logic is like this:

iterate the list, and install when `item` is `ppa:certbot/certbot' and bbb_letsencrypt_enable` or not `ppa:certbot/certbot`

Ok then this makes sense.

BrutalBirdie (approved):

tasks/repositories.yml
crosscodr: Yes, it works out. I have this version in production. Thanks for approving!
Unfortunately, there is no nicer way to do this with ansible. At least, i don't know one.
I thought about another option. The simplest is to write an extra task. But this is rather verbose.
I liked this better.

Davka (approved):

PR #149, opened 2021-06-03
add backend for html5-client according to https://docs.bigbluebutton.…

einhirn:
add backend for html5-client according to https://docs.bigbluebutton.org/2.2/customize.html#collect-feedback-from-the-users

With this, people can add 
```yaml
bbb_meteor:
  public:
    app:
      askForFeedbackOnLogout: true
    clientLog:
      external:
        enabled: true
        url: "https://{{ bbb_hostname }}/html5log"
```
to their config and the logs get written to `/var/log/nginx/html5-client.log`

When switching the variable "bbb_client_log_enable" to "false", the changes are reverted - only the log file is left.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | `bbb_allow_mail_notifications`  | Set this to true if you want GreenLight to send verification emails upon the creation of a new account | `true` |
 | `bbb_disable_recordings` | Disable options in gui to have recordings | `no` | [Recordings are running constantly in background](https://github.com/bigbluebutton/bigbluebutton/issues/9202) which is relevant as privacy relevant user data is stored |
 | `bbb_api_demos_enable` | enable installation of the api demos | `no` | |
+| `bbb_client_log_enable` | enable installation of the nginx-full and config for client logging according to [BBB Customization Docs](https://docs.bigbluebutton.org/2.2/customize.html#collect-feedback-from-the-users). See "METEOR" Section below for needed `bbb_meteor` values. | `false` | |
 | `bbb_mute_on_start:` | start with muted mic on join | `no` | |
 | `bbb_app_log_level:` | set bigbluebutton log level | `DEBUG` | |
 | `bbb_meteor:` | overwrite settings in meteor | `{}` | |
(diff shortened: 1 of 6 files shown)


BrutalBirdie:
Please be sure ansible-lint does not produce any errors.
This might solved by merging the latest changes from the `bbb/2.3` branch.

crosscodr (commented):
Looks good to me, but I don't see this part of the doc:
```
yq w -i $HTML5_CONFIG public.clientLog.external.enabled true
yq w -i $HTML5_CONFIG public.clientLog.external.url     "$PROTOCOL://$HOST/html5log"
yq w -i $HTML5_CONFIG public.app.askForFeedbackOnLogout true
```

A comment in README.md would be nice, too :)

einhirn:
> Looks good to me, but I don't see this part of the doc:

yes, I only mentioned it here in the original comment, which I just updated with the last line: https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/149#issue-660787738

> A comment in README.md would be nice, too :)

Oh, of course, I'll whip something up...

crosscodr:
> > Looks good to me, but I don't see this part of the doc:
> 
> yes, I only mentioned it here in the original comment, which I just updated with the last line: [#149 (comment)](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/149#issue-660787738)
Oh yes, I overlooked that afterwards.
Thanks for the README-Entry

BrutalBirdie:
Instead of using `files/` I would suggest using template.
This way later on down the road these files can be templated if needed.

---
After reading the code again . . . 🤔
Since you do `copy` and `remove` based on `bbb_client_log_enable` template might not be the right option here... or?

Suggestion.
Yes use templates and only on the `remove` steps use [file](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/file_module.html):
```yaml
- name: Remove file (delete file)
  ansible.builtin.file:
    path: /etc/foo.txt
    state: absent
```
This way the templates will be created based on `bbb_client_log_enable` and deleted.
Your thoughts?

crosscodr:
> Instead of using `files/` I would suggest using template.
> This way later on down the road these files can be templated if needed.

> Your thoughts?

Hi @BrutalBirdie, I would say, if it is needed to template a file, one could change the copy task to a template task and move from files to templates. Just my two cents :)

einhirn:
Template parsing takes time, so why do it if there aren't any variables in the file?
@crosscodr - exactly.

BrutalBirdie (approved):

BrutalBirdie:
NOTE:
I did not test this yet, reminder to myself do that :D

Davka (approved):

README.md
```diff
@@ -312,6 +313,19 @@ bbb_meteor:
           viewer: 8
 ```
```
crosscodr, line 315:
```suggestion
#### User Feedback logging
```

crosscodr (changes requested):

crosscodr (approved):

PR #152, opened 2021-06-04
configures the port for stun-server

### LINKED ISSUE #145: bbb_stun_servers does not set the port
You could set the port via the URL, but that's not so nice. It would be better if you could give the port.

Davka:
fixes #145

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -19,6 +19,7 @@ bbb_coturn_max_port: 65535
 bbb_turn_enable: true
 bbb_stun_servers:
   - server: "{{ bbb_hostname }}"
+    port: " {{ bbb_coturn_port_tls }}"
 bbb_ice_servers: []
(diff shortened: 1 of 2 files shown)


BrutalBirdie (approved):

n0emis:
Hmm, I am not sure on this change. If merged this definitely would need a major release since it is a rather important breaking change. Therefor I am asking myself if this change really is necesarry? Since there is no place where just the hostname of the TURN-Server is required I see no necessity to change this.

Davka:
> Hmm, I am not sure on this change. If merged this definitely would need a major release since it is a rather important breaking change. Therefor I am asking myself if this change really is necesarry? Since there is no place where just the hostname of the TURN-Server is required I see no necessity to change this.

Ok, you are right. It also works if you include the port in the URL.
- server: "url: 443"

PR #153, opened 2021-06-04
add sip ufw rule

### LINKED ISSUE #143: UFW and SIP-Dial-In
If the UFW is activated, the SIP dial-in is blocked after a while

Davka:
fixes #143

### DIFF
--- a/tasks/firewall/ufw_enable.yml
+++ b/tasks/firewall/ufw_enable.yml
@@ -75,6 +75,13 @@
       port: ssh
       proto: tcp
 
+  - name: allow sip-proxy connection
+    community.general.ufw:
+      rule: allow
+      from: '{{ bbb_dialin_provider_ip }}'
+      to_port: '5060'
+    when: bbb_dialin_enabled | bool and bbb_dialin_provider_ip is defined
+
   - name: Setup default policy
     community.general.ufw:
       policy: deny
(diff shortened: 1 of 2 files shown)


BrutalBirdie (approved):

tasks/firewall/ufw_enable.yml
```diff
@@ -75,6 +75,13 @@
       port: ssh
       proto: tcp
 
+  - name: allow sip-proxy connection
+    community.general.ufw:
+      rule: allow
+      from: '{{ bbb_dialin_provider_ip }}'
+      to_port: '5060'
+    when: bbb_dialin_enabled | bool and bbb_dialin_provider_ip is defined
```
crosscodr, line 83: I think the `is defined` check is not sufficient here, because you define it in defaults.main.yml.

```suggestion
    when: bbb_dialin_enabled | bool and bbb_dialin_provider_ip is defined and bbb_dialin_provider_ip | length > 0
```
To do it more correct, one could use this one:
```suggestion
    when: bbb_dialin_enabled | bool and bbb_dialin_provider_ip | ansible.netcommon.ipaddr()
```
But then, the admin who runs ansible need to install netaddr, see
https://docs.ansible.com/ansible/2.9/user_guide/playbooks_filters_ipaddr.html

crosscodr (changes requested):

crosscodr (approved):

PR #159, opened 2021-06-07
set TasksMax=infinity in kurento-x.services

Davka:


crosscodr:
This is a upstream change from
https://github.com/bigbluebutton/bigbluebutton/blob/develop/bigbluebutton-config/bin/apply-lib.sh#L107

crosscodr (approved):

PR #161, opened 2021-06-07
set LimitNOFILE depending on cpu-count

Davka:


crosscodr (approved):

crosscodr:
Follow https://github.com/Davka/bbb-install/blob/master/bbb-install.sh#L666

BrutalBirdie (approved):

PR #165, opened 2021-06-08
Add always tag to var checking and registering tasks

ChrisKar96:
Add always tag to var checking and registering tasks since the values are needed for other tasks to run successfully.

So, even though #163 is ok, if you run a tag that requires a variable, without also running `runtimevars`, it will fail. Also, `checkvars` needs to be called to check if you have incorrect configuration.
This is why I add the always tag to those tasks, so the required variables will be available for other tasks. 

I wanted to always run `check-for-sessions` (added in #140) as well, but there is a problem there. It needs the `bbb_secret` variable, which is gathered by `runtimevars`. I don't know if we should move the `runtimevars` task before `check-for-sessions`, because that will put it before `installation` as well, likely making the role fail for first time installations. That said, was #140 checked in a fresh install? I think it will fail for `bbb_secret` not being available, but I might be wrong if is defined/ignored somewhere and I haven't noticed. 

Anyway, a solution would be to create an extra task to determine if bbb is already installed and if true we call `runtimevars` right after `checkvars`, else we run it after installation, as is now, and skip `check-for-sessions`. If you can think of a better solution please share it, otherwise I can start implementing the one I proposed.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -3,6 +3,7 @@
 - import_tasks: checkvars.yml
   tags:
     - checkvars
+    - always
 
 - include_tasks: check-for-sessions.yml
   when: bbb_check_for_running_meetings
@@ -38,6 +39,7 @@
 - import_tasks: runtimevars.yml
   tags:
     - runtimevars
+    - always
 
 - import_tasks: letsencrypt.yml
   when: bbb_letsencrypt_enable | bool and not bbbcert.stat.exists
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -5,10 +5,21 @@
     - checkvars
     - always
 
+- import_tasks: runtimevars.yml
+  tags:
+    - runtimevars
+    - always
+  when:
+    - not bbb_secret is defined or not bbb_baseurl is defined or not bbbcert is defined
+  ignore_errors: true
+
 - include_tasks: check-for-sessions.yml
-  when: bbb_check_for_running_meetings
+  when:
+    - bbb_check_for_running_meetings
+    - bbb_secret is defined
   tags:
     - check-for-sessions
+    - always
 
 - import_tasks: firewall/ufw_enable.yml
   when: bbb_firewall_ufw['enabled']
@@ -40,6 +51,8 @@
   tags:
     - runtimevars
     - always
+  when:
+    - not bbb_secret is defined or not bbb_baseurl is defined or not bbbcert is defined
 
 - import_tasks: letsencrypt.yml
   when: bbb_letsencrypt_enable | bool and not bbbcert.stat.exists
--- a/tasks/runtimevars.yml
+++ b/tasks/runtimevars.yml
@@ -4,6 +4,7 @@
   stat:
     path: "{{ bbb_ssl_cert }}"
   register: bbbcert
+  when: not bbbcert is defined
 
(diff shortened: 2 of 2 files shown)


crosscodr:
> Since bbb_secret is cached in runtimevars.yml, it should work after the first run.

Maybe the caching is not used per default, so we need to do more about it.

ChrisKar96:
Actually, since adding the tags in main, the always tags inside ​`check-for-sessions.yml​` individual tasks don't work, since not including `check-for-sessions` tag will result in the yml not being included as a whole. So, those should be removed in any case.

BrutalBirdie (approved):

ChrisKar96:
My last commit Fixes #168, being a permanent replacement for #169 that just disabled `check-for-sessions.yml` from ever running because `bbb_secret` was not defined at this point unless manually hard-coded in the playbook variables by the user.



> 
> 
> Actually, since adding the tags in main, the always tags inside ​`check-for-sessions.yml​` individual tasks don't work, since not including `check-for-sessions` tag will result in the yml not being included as a whole. So, those should be removed in any case.

Turns out these can not be removed because an error is thrown if they are removed.

Let me know if anything else should be changed before this is merged.

BrutalBirdie (approved):

crosscodr (approved):

einhirn:
The conditionals on the second run of `runtimevars` make ansible skip those tasks in my case. But "bbbcert" also gets overwritten with `        "skip_reason": "Conditional result was False",` so "certbot.yml" fails because `bbbcert.stat` is not defined anymore...

ChrisKar96:
For ansible to skip the second run of `runtimevars`, it means you tried to rerun the playbook on an already configured host, right? 

I didn't have a problem, but I am using ansible 2.10, maybe different versions behave differently on this part, I am not sure, but just in case what version did you use?

When I tested this PR, initially I included an `and bbbcert.stat.exists` check that worked perfectly for the first run of `runtimevars` but failed the second time with an error that was something like `stat doesn't exist for dict type object`, so I removed that ckeck. For debugging I printed the var right before the second run of `runtimevars` and `bbbcert` was there with the result of the previous run of the task, as it should be. Maybe ansible initializes the var again when the task runs and doesn't keep the previous value? We could probably circumvent this with another var, initialized in a different set of tasks that would hold the actual needed value of bbbcert.stat, but I need to test it and I am on vacation now.

I can try and assist if someone is willing to tackle this, or I will look into it after Monday.

einhirn:
> For ansible to skip the second run of `runtimevars`, it means you tried to rerun the playbook on an already configured host, right?

Correct.

> I didn't have a problem, but I am using ansible 2.10, maybe different versions behave differently on this part, I am not sure, but just in case what version did you use?

I'm also using 2.10(.2 to be exact.)

> I can try and assist if someone is willing to tackle this, or I will look into it after Monday.

@crosscodr and I already have a solution, thanks. We just `include_tasks` instead of `import_tasks`. The difference is where `when` is applied: at each imported task vs. only once for the whole include. That prevents the task itself to be skipped and a "skip_reason" overwriting the original content of the var.

ChrisKar96:
Nice! Should be an easy fix then, I wasn't familiar with the difference between the two, so thanks for explaining.

PR #174, opened 2021-06-10
Add nginx sources and debug info

DerDanilo:
- Add nginx upstream sources to use latest stable (mainline) nginx packages
- Add nginx TLS 1.3 support
- Add debug info message before installing bbb packages and dependencies since this step takes quiet a while and the playbook may seem stuck

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,7 +6,7 @@ bbb_check_for_running_meetings: true
 bbb_letsencrypt_enable: true
 bbb_letsencrypt_api: https://acme-v02.api.letsencrypt.org/directory
 bbb_nginx_install_from_source: true
-nginx_branch: 'mainline'
+bbb_nginx_branch: 'mainline'
 bbb_nginx_privacy: true
 bbb_nginx_listen_https: true
 bbb_nginx_root: /var/www/bigbluebutton-default
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,7 +6,7 @@ bbb_check_for_running_meetings: true
 bbb_letsencrypt_enable: true
 bbb_letsencrypt_api: https://acme-v02.api.letsencrypt.org/directory
 bbb_nginx_install_from_source: true
-bbb_nginx_branch: 'mainline'
+bbb_nginx_use_mainline_branch: false
 bbb_nginx_privacy: true
 bbb_nginx_listen_https: true
 bbb_nginx_root: /var/www/bigbluebutton-default
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,6 +6,7 @@ bbb_check_for_running_meetings: true
 bbb_letsencrypt_enable: true
 bbb_letsencrypt_api: https://acme-v02.api.letsencrypt.org/directory
 bbb_nginx_install_from_source: true
+# If this is set to 'false', the current stable branch will be used instead.
 bbb_nginx_use_mainline_branch: false
 bbb_nginx_privacy: true
 bbb_nginx_listen_https: true
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -5,6 +5,8 @@ bbb_upgrade_packages: "{{ 'yes' if bbb_state == 'latest' else 'no' }}"
 bbb_check_for_running_meetings: true
 bbb_letsencrypt_enable: true
 bbb_letsencrypt_api: https://acme-v02.api.letsencrypt.org/directory
+bbb_nginx_install_from_source: true
+nginx_branch: 'mainline'
 bbb_nginx_privacy: true
 bbb_nginx_listen_https: true
 bbb_nginx_root: /var/www/bigbluebutton-default
--- a/tasks/installation.yml
+++ b/tasks/installation.yml
@@ -13,6 +13,28 @@
     dest: /etc/alternatives/java
     state: link
 
+- name: Add NGINX sources
+  become: yes
+  block:
+    - name: "Install NGINX sources dependencies"
+      package: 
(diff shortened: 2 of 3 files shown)


defaults/main.yml
```diff
@@ -5,6 +5,8 @@ bbb_upgrade_packages: "{{ 'yes' if bbb_state == 'latest' else 'no' }}"
 bbb_check_for_running_meetings: true
 bbb_letsencrypt_enable: true
 bbb_letsencrypt_api: https://acme-v02.api.letsencrypt.org/directory
+bbb_nginx_install_from_source: true
+bbb_nginx_branch: 'mainline'
```
crosscodr, line 9: I would rather like to default this to 'stable', since the current stable version (1.20) is already much newer than bionic default (1.14) and also supports TLS 1.3.

```suggestion
bbb_nginx_branch: 'stable'
```

templates/nginx/vhost.conf
```diff
@@ -13,7 +13,7 @@ server {
     ssl_certificate_key {{ bbb_ssl_key }};
     ssl_session_cache shared:SSL:10m;
     ssl_session_timeout 1d;
-    ssl_protocols TLSv1.2;
+    ssl_protocols TLSv1.2 TLSv1.3;
```
crosscodr, line 16: When the use of the official nginx-repo is selectable, one should make 'TLSv1.3' depend on that choice.

crosscodr (changes requested):
Hi @DerDanilo,

thank you for your PR!
I have put some comments below and would ask you to put a note on this in the README.md.

defaults/main.yml
DerDanilo: One may go with the stable release but the mainline release is very stable as well. You may write any other word into 'bbb_nginx_branch' and the "stable" package will be deployed. 

`deb https://nginx.org/packages/{{ (bbb_nginx_branch == 'mainline') | ternary('mainline/', '') }}`

DerDanilo: I added a switch to enable mainline (if required)

templates/nginx/vhost.conf
DerDanilo: TLSv1.3 support is included in Nginx since version 1.13. I don't understand why TLS 1.3 was not enabled by default in the BBB templates. So I think that this should be the default anyways. 

> "Nginx version 1.13.0 or greater built against OpenSSL 1.1.1 or greater."

defaults/main.yml
DerDanilo: It seems there is no 'stable' version available via the official nginx repo. Adding a community ppa might be our only option. But then again this seems a bit overhead. What do you think? Which ppa to use then?

crosscodr: Yes there is a official stable repo:
https://nginx.org/packages/ubuntu/

templates/nginx/vhost.conf
crosscodr: Oh yes, then we could even prefer TLS1.3 over TLS1.2:
```suggestion
    ssl_protocols TLSv1.3 TLSv1.2;
```

Davka: > TLSv1.3 support is included in Nginx since version 1.13. I don't understand why TLS 1.3 was not enabled by default in the BBB templates. So I think that this should be the default anyways.
> 
> > "Nginx version 1.13.0 or greater built against OpenSSL 1.1.1 or greater."

I think historically. In ubuntu 16, nginx was not compiled against openssl 1.1.1.

crosscodr: Maybe, but now, we are on Ubuntu 18.04.
I tested it. Even with ubuntu-packaged nginx (1.14) you can use TLSv1.3, since openssl 1.1.1 is installed on Ubuntu 18.04

crosscodr (changes requested):

defaults/main.yml
crosscodr: See also https://nginx.org/en/linux_packages.html#Ubuntu

DerDanilo:
Added the suggested. Please check and merge

crosscodr:
Hi @DerDanilo, please resolve the ansible-lint failures. Rest looks fine!

DerDanilo:
Hope that I got it now. 
There seem to be some other lint erros regarding other playbooks as well.

tasks/installation.yml
```diff
@@ -13,11 +13,31 @@
     dest: /etc/alternatives/java
     state: link
 
+- name: Add NGINX sources
+  become: true
+  block:
+    - name: "Install NGINX sources dependencies"
+      apt:
+        name: "software-properties-common"
+        state: latest
```
crosscodr, line 22:
```suggestion
        state: "{{ bbb_state }}"
```

defaults/main.yml
```diff
@@ -5,6 +5,9 @@ bbb_upgrade_packages: "{{ 'yes' if bbb_state == 'latest' else 'no' }}"
 bbb_check_for_running_meetings: true
 bbb_letsencrypt_enable: true
 bbb_letsencrypt_api: https://acme-v02.api.letsencrypt.org/directory
+bbb_nginx_install_from_source: true
```
crosscodr, line 8: @Davka, @BrutalBirdie should we change default source for nginx?
@DerDanilo The variable name `bbb_nginx_install_from_source` is a bit misleading, in my opinion.
```suggestion
bbb_nginx_install_from_official_repo: true
```

crosscodr (changes requested):

BrutalBirdie:
closed due to inactivity.
Please send a message if you wish to work on this PR again.

PR #78, opened 2021-01-03
fix travis

juju4:
manual install of wheel to avoid "error: invalid command 'bdist_wheel'"

PR #182, opened 2021-06-13
Add Github action workflow - full ansible run

juju4:
This is an adaptation of previous travis-ci config for github (https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/78), but directly on github host virtual environments.
Simpler to read but possible interactions/conflict with the many developer tools added.
I have an alternate config with kitchen+lxd which is cleaner for base image but I have concern on maintainability of the kitchen lxd_cli plugin which force an old 1.x kitchen version. (https://github.com/juju4/ansible-role-bigbluebutton/actions?query=branch%3Adevel-23)

schedule is not enabled, to validate.
focal is configured but marked as experimental, meaning fail is not impacting

few issues pending
- ufw settings seems to have issue and currently disabled
- turn does not seem to start correctly. temporary ignored in verify.yml

minor changes
- removing unused variables, test suites
- restrict all molecule to hosts group bbb
- apt key download retries
- use filter comment for ansible_managed as required if using a multiline one


juju4:
Unclear on why PR CI failing on bbb packages install (https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/182/checks?check_run_id=2814486088#step:6:21496) but succeeding in my tree (https://github.com/juju4/ansible-role-bigbluebutton/actions?query=branch%3Adevel-23b)

defaults/main.yml
```diff
@@ -205,7 +205,7 @@ bbb_ufw_allow_networks: "{{ bbb_ufw_allow_networks_custom | default([]) + bbb_uf
 bbb_ufw_reject_networks: "{{ bbb_ufw_reject_networks_custom | default([]) + bbb_ufw_reject_networks_default }}"
 bbb_firewall_ufw:
   # enable/disable the ufw
-  enabled: true
+  enabled: false
```
crosscodr, line 208: The testrun should not change this default.

BrutalBirdie: Good catch

juju4: as said above, it has been creating issue on my side (dns/system-resolved).
unclear why because from my reading, there are no outbound firewall rules but.

reverted

juju4: Following error when ufw enabled
https://github.com/juju4/ansible-role-bigbluebutton/runs/2823828866?check_suite_focus=true#step:6:3378
```
Failed to download key at https://ubuntu.bigbluebutton.org/repo/bigbluebutton.asc: Request failed: <urlopen error [Errno -3] Temporary failure in name resolution>
```

after failure info
https://github.com/juju4/ansible-role-bigbluebutton/runs/2823828866?check_suite_focus=true#step:9:241

EdJoPaTo (commented):
Some nitpicks I found when glancing upon it

.github/workflows/default.yml
```diff
@@ -0,0 +1,155 @@
+---
+name: AnsibleCI
```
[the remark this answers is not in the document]
juju4: default comes from test-kitchen usual default test suite/folder
filename can be AnsibleCI, test-install or whatever

EdJoPaTo:
In 5c31362 and 958bd0ce81df1ac6eb05ef5ee9f6517ef3f710b5 the `working-directory` was removed.
I am curious what the problem there was. I looked at the action logs but the errors seem independent from that. Am I missing something there?
(Don't want to judge, just wanna learn more)

juju4:
In either case below, those were the only changes and cause the CI to fail early, so reverted until identify how to make it work
https://github.com/juju4/ansible-role-bigbluebutton/runs/2864960066?check_suite_focus=true#step:6:12
https://github.com/juju4/ansible-role-bigbluebutton/runs/2864905894?check_suite_focus=true#step:4:20

it seems there are limitations attached to variable use
for example, from https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#env-context
> The env context syntax allows you to use the value of an environment variable in your workflow file. You can use the env context in the value of any key in a step except for the id and uses keys. For more information on the step syntax, see "Workflow syntax for GitHub Actions."

trying without variable

juju4:
same, not working => reverting
https://github.com/juju4/ansible-role-bigbluebutton/runs/2866631992?check_suite_focus=true#step:6:11

.github/workflows/default.yml
```diff
@@ -44,7 +43,7 @@ jobs:
           ls -l /opt/hostedtoolcache/Python/3.9.1/x64/bin || true
           echo "/opt/hostedtoolcache/Python/3.9.1/x64/bin" >> $GITHUB_PATH
           ansible --version
-          cd $GITHUB_WORKSPACE/${{ env.ROLEPATH }}
+          cd $GITHUB_WORKSPACE/ebba.bigbluebutton
```
EdJoPaTo: I guess the reason here could also be the typo
```diff
-ebba
+ebbba
```

juju4: yes, typo. but consistent with checkout path so unlikely to be the root cause.
fixed.

juju4:
Solved firewall issue by commenting out 10.0.0.0/8 from ufw rejected network as it is the subnet used github CI. Private network exclusion is really architecture/environment dependent so I think it would better to allow also 192.168/16. Bogon networks filtering is usually better done at edge.

+greenlight pause as sometimes database not coming up fast enough
+refine a bit molecule verify.

Need fix but likely in a separate PR
- idempotence
- some molecule/verify tasks failing: Ensure red5 config file is present, Ensure turnserver process is running+ports, Validate services started

.github/workflows/default.yml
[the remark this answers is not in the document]
juju4: what should I go here? is AnsibleCI fine?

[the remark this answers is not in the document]
EdJoPaTo: AnsibleCI in the CI environment for a git Repo for an Ansible sounds not that helpful…
Sadly I don't know a good name here…

Basically it tests the whole installation? end to end? integration test?

BrutalBirdie:
@juju4 
I am so sorry this took so long but I am finally reviewing this PR for good.
Since there where so many changes and breaks with this PR I am currently fixing them all.

as can be followed here: https://github.com/ebbba-org/ansible-role-bigbluebutton/actions/workflows/default.yml
and here: https://github.com/ebbba-org/ansible-role-bigbluebutton/compare/juju4-devel-23b

So I will close this PR and repopen it with this branch and then finally merge it when its good.

🍻

BrutalBirdie:
moved to #269

PR #184, opened 2021-06-14
fix check-for-sessions: since the uri task doesn't fail anymore, the …

einhirn:
since the uri task doesn't fail anymore, the following check had to be adjusted.

### DIFF
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -8,6 +8,11 @@
     changed_when: false
     tags: always
 
+  - name: debug
+    debug:
+      var: bbb_getmeetings
+    tags: always
+
   - block:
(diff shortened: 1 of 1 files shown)


crosscodr (approved):

BrutalBirdie (approved):

crosscodr:
Hi @einhirn, I approved your changes, but just saw, that the debug-task prints the whole output of `getMeetings` which is not necessary in my opinion.
Can you revert this? @n0emis, @BrutalBirdie, what do you think?

einhirn:
Gna - sorry, I didn't clean up before hastily creating that PR... Of course that debug task is not neccessary. Since a revert is a new Pull Request undoing everything, I'd rather create a new one to remove the debug task, if that's ok.

PR #186, opened 2021-06-15
fix runtimevars

crosscodr:
Use include instead of import, since imported tasks override registerd
variables from the first run.
See: https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/165#issuecomment-861393376

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -5,7 +5,12 @@
     - checkvars
     - always
 
-- include_tasks: runtimevars.yml
+- include_tasks:
+    file: runtimevars.yml
+    apply:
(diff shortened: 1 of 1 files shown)


Davka (approved):

crosscodr:
I added a commit which applies the tags in tasks/main.yml to the tasks of runtimevars.yml, acording to 
https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#tag-inheritance-for-includes-blocks-and-the-apply-keyword.

BrutalBirdie (approved):

PR #188, opened 2021-06-15
improve ansible tagging

### LINKED ISSUE #181: Make dependency roles optional
Please make it optional to use the hard dependencies for docker and nodejs.  If one wants to deploy the two components otherwise this might create conflicts.

There could be booleans that handles this:
```

bbb_include_role_geerlingguy_nodejs: true
bbb_include_role_geerlingguy_docker: true
```

Default should of course be 'true'.

crosscodr:
Improves and adds some additional tags.
Fixes use of `greenlight` and `monitoring` tags, by adding the tags to the depended docker tasks.
Also fixes #181.


meta/main.yml
```diff
@@ -22,4 +22,8 @@ dependencies:
   - role: geerlingguy.nodejs
     vars:
       nodejs_version: "{{ bbb_nodejs_version }}"
+    tags: geerlingguy_nodejs
```
einhirn, line 25: An additional tag `geerlingguy` would help shorten the commandline when I want to skip both dependencies.

einhirn (changes requested):
An additional tag "geerlingguy" for both roles would reduce command line length when I want to skip both dependencies in one go.

Davka (approved):

crosscodr:
Hi @BrutalBirdie, I resolved the merge conflicts. What do you think?

BrutalBirdie (approved):

PR #194, opened 2021-06-17
fix: removed 'is' from conditonal

Nemental:
```
The conditional check ''<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content' failed. The error was: template error while templating string: expected token 'end of statement block', got '.'. String: {% if '<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content %} True {% else %} False {% endif %}
```

### DIFF
--- a/tasks/check-for-sessions.yml
+++ b/tasks/check-for-sessions.yml
@@ -16,5 +16,5 @@
     when:
       - not ansible_check_mode
       - bbb_getmeetings.status == 200
-      - "'<messageKey>noMeetings</messageKey>' is not in bbb_getmeetings.content"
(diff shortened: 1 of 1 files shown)


crosscodr:
Your fix seems legit. However, I don't get this error-message. Which version of ansible do you run?
Mine is ansible 2.9.21.

BrutalBirdie:
I running ansible `2.11.1` and also never encountered this issue.

Nemental:
I don't really get this error message too 😄 My current installed version is `2.9.16`. Apart from the error message, the pull request corrects the syntax.

crosscodr (approved):

BrutalBirdie (approved):

PR #201, opened 2021-06-22
Fix issues with #95 (default kurento listen interface)

ChrisKar96:
Fix issues with #95 (default kurento listen interface) to be merged upstream


crosscodr:
Hi @ChrisKar96, thank you for your PR!
I would suggest to move the tasks from tasks/config.yml to tasks/kurento.yml.
There are already tasks, that change something in `/etc/kurento/modules/kurento/WebRtcEndpoint.conf.ini`.

ellakdev:
I agree with your suggestion, I just copied what worked for me but it can definitely be improved. I will update shortly.

ChrisKar96:
I agree with your suggestion, I just copied what worked for me but it can definitely be improved. I will update shortly.

crosscodr:
Looks good to me.
However, I'm not sure if we should change the default, by defining `bbb_kurento_interfaces`in defauts/main.yml.
@BrutalBirdie @n0emis, what do you think?

ChrisKar96:
My opinion is that changing the default will fix more than it breaks (if it even breaks something). Kurento straight up wasn't working for me before this change. Furthermore, should anyone experience breaks after the merge, they can override the variable in their configuration to restore the original behavior. 

Of course, the same is true the other way around, we can remove the default and anyone can define the variable in their config if they experience problems. 

I still prefer keeping the default, but in either case, maybe we should expand a little more in README on how and when to use the variable.

Let me all know what you think.

crosscodr (approved):

Davka (approved):

PR #205, opened 2021-06-28
Added github action for generating changelog and release

BrutalBirdie:


.github/workflows/changelog_builder.yml
```diff
@@ -0,0 +1,25 @@
+name: 'CI'
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  release:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-18.04-LTS
```
EdJoPaTo, line 10: Is there any need to pin this to a specific ubuntu version?
I think just using ubuntu-latest should be fine here and will not fail when the github runners for ubuntu 18.04 are gone

PR #215, opened 2021-07-20
move logback.xml to files; use indentation following upstream

crosscodr:
This commit makes two changes:
1. Move logback.xml from `templates` to `files`, since it contains no jinja elements.
2. Use tabs of size 4 as upstream does. Our own changes are now much more visible in ansible diff.


crosscodr:
Ansible diff (ansible-playbook called with --diff or -D) still does not show a good diff :(
I think this is due to the fact, that the whole file is overwritten by the task.
The changes are sane, though.

BrutalBirdie (approved):
lgtm.

Davka (approved):

PR #216, opened 2021-07-20
Bbb/2.3 monitoring

Davka:
This PR extends the monitoring to include the option of controlling the IP rules instead of htpasswd, and some adjustments have been made


README.md
```diff
@@ -83,6 +83,11 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
...
 | `bbb_monitoring_external` | Enable exposure to nginx | `false` | Can be reached under `/mon/bbb` and `/mon/node` - requires `htpasswd` and `htpasswd_user`. If `bbb_monitoring_systemd_enable` is enabled, no Node Exporter installation process is included |
+| `bbb_monitoring_external_metrics_port` | Port of bbb-exporter | `9688` |  default port 9866 is defined by the exporter [itself](https://github.com/greenstatic/bigbluebutton-exporter/blob/master/bbb-exporter/settings.py#L37) |
+| `bbb_monitoring_external_node_port` | Port of node-exporter | `9100` | default port 9100 is by the bbb-exporter [itself](https://github.com/greenstatic/bigbluebutton-exporter/blob/master/extras/node_exporter/docker-compose.yaml#L10) |
+| `bbb_monitoring_use_ip_restriction` | Restricting access by IP Address instead using htpasswd | `false` | |
+| `bbb_monitoring_use_restriction_combination`| Use combination of IP Address restriction and htpasswd | `false` | |
+| `bbb_monitoring_external_ip_list` | A list of ip addresses for restricting access | `[]` | in array with key `address` |
```
ChrisKar96, line 90: I would suggest this variable to be renamed 'bbb_monitoring_external_ip_whitelist' for clarity. It's not unclear now, but the change would make it even better in my opinion.

Davka: The variable has been renamed

tasks/monitoring/all_in_one.yml
```diff
@@ -13,11 +13,18 @@
...
-    - prometheus.yaml
   notify: restart monitoring container
 
+- name: copy prometheus files into place
+  template:
+    src: "monitoring/prometheus.yaml.j2"
+    dest: "{{ bbb_monitoring_all_in_one_directory }}/prometheus.yaml"
+    mode: '0644'
+  notify: restart monitoring container
+  when: bbb_monitoring_all_in_one_enable | bool and not bbb_monitoring_external | bool
```
crosscodr, line 24: In templates/monitoring/docker-compose.yaml.j2 the prometheus-container is still deployed.
This should be done with the same logic as here.

tasks/monitoring/main.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - import_tasks: all_in_one.yml
-  when: bbb_monitoring_all_in_one_enable | bool
+  when: bbb_monitoring_all_in_one_enable | bool or bbb_monitoring_external | bool
```
crosscodr, line 3: Maybe it's a good idea to splitout the external tasks further and include them separately when bbb_monitoring_external | bool or bbb_monitoring_all_in_one_enable | bool

crosscodr (changes requested):

Issue #220, opened 2021-07-23
ufw rules are not reset after rerunnning the play


d3hof:
**Describe the bug**
ufw rules are not reset when rerunning the play as mentioned here #217 

**To Reproduce**
1. Run the playbook once with coturn_enabled: true
2. Rerun it again with coturn_enabled: false
3. Coturn ufw rules are still there

**Expected behavior**
ufw rules should be reset with every run

**Ansible version**
  ansible 2.10.12
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/d3h/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  executable location = /usr/bin/ansible
  python version = 3.8.10 (default, Jun  2 2021, 10:49:15) [GCC 9.4.0]

BrutalBirdie:
we should remove greenlight and coturn from this role anyway.
This will be the task to fix this issue.
Closing this issue for now.

PR #164, opened 2021-06-08
Improve UFW configurability

ChrisKar96:
Add variables to extend UFW allowed/rejected networks and configure different ssh port

PR #221, opened 2021-07-23
reset firewall rules

d3hof:
fixes this https://github.com/ebbba-org/ansible-role-bigbluebutton/issues/220

### DIFF
--- a/tasks/firewall/ufw_enable.yml
+++ b/tasks/firewall/ufw_enable.yml
@@ -13,6 +13,10 @@
       state: enabled
       policy: allow
 
+  - name: reset ufw rules
+    community.general.ufw:
+      state: reset
+
   - name: Set logging
     community.general.ufw:
       logging: 'on'


crosscodr:
Hi @d3hof, thank you for this PR.
Since the ufw tasks are not idempotent right now, this patch does not make it worse in this sense.
In the long run, i would prefer a idempotent firewall solution, maybe with templates.
I wonder if there is a good solution to template /etc/ufw/user.conf (this are essentially iptables rules).

d3hof:
You're right it's just a cheap solution for a problem I encountered running the role. 
I could look into templating the rules, but it might take some time ;)

crosscodr (approved):

crosscodr:
> You're right it's just a cheap solution for a problem I encountered running the role.
> I could look into templating the rules, but it might take some time ;)

Yeah this would only be possible, if the whole ufw-tasks are rewritten.
In the future I would prefer nftables. Maybe when bbb runs on Ubuntu 20.04 or above and docker supports nftables a bit better.

tasks/firewall/ufw_enable.yml
```diff
@@ -13,6 +13,10 @@
       state: enabled
       policy: allow
 
+  - name: reset ufw rules
+    community.general.ufw:
+      state: reset
+
```
BrutalBirdie, line 19: Quite brute, but it works :D

crosscodr: From the man page of ufw:
> **reset**  Disables and resets firewall to installation defaults.

This will allow anything. I tested it. All policies are set to ACCEPT and all rules are set to "ACCEPT anywhere anywhere".
So this task could even replace the task above (policy allow). But we need an activate task afterwards. The reset task disables the whole firewall. Adding rules after does not activate the firewall again.

`ufw reset` also makes backups of the current rules. This could mess up /etc/ufw after a while.

How can i revert my approvement? :D

d3hof:
According to my tests the default incoming rules are not reset. I almost locked myself out from a server because the ufw default deny wasn't removed.

crosscodr:
> According to my tests the default incoming rules are not reset. I almost locked myself out from a server because the ufw default deny wasn't removed.

At least the firewall is inactive after a ansible-run with this patch. This has to be fixed.

tasks/firewall/ufw_enable.yml
ChrisKar96: I researched this quite extensively when I was working on #164 
There is no perfect way to do this, at least at the moment.

The suggested course of action is, as @crosscodr said, to move the `- name: Activate ufw` task at the bottom of the file, so that reset happens -> rules added -> firewall activated

I guess a different approach would be to parse the number of rules from `ufw status numbered` and then delete them all in a loop before adding them again, but I don't like this either because it will be displayed as changed state even when changing nothing.

d3hof: The point is if the default deny is set before you get locked out, that's why you have actually allow default, reset it and then turn it on again. I'll check out if it's possible to somehow fix the stuff that happens in /etc/ufw

ChrisKar96: Is there any reason we should stick with UFW? I know it's more work, but I think creating a template for iptables and restoring it on the host, might be the way to go. We could use [this](https://docs.ansible.com/ansible/2.10/collections/community/general/iptables_state_module.html) and the template would include only the rules required by the enabled features, controlled by ansible variables (ex not including coturn rules if coturn not enabled). We could also use the [built-in iptables module](https://docs.ansible.com/ansible/2.10/collections/ansible/builtin/iptables_module.html) to directly manipulate the rules, but I prefer the template idea. I could help creating the template, if we decide migrating away from UFW is a good option.

crosscodr: Yeah, I would also prefer direct use of iptables-templates :)

crosscodr: I can write something up till the end of the week.

d3hof: We ran into some issues running bigbluebutton with iptables, but if you can make it work I am ok with that :)

ChrisKar96: > 
> 
> I can write something up till the end of the week.

Feel free to ping me and let me know how I can assist.

crosscodr: @d3hof ufw is just a wrapper around iptables, so if ufw works with bbb, iptables will do also :)
@ChrisKar96, thank you. I'll put up some WIP-PR soon.

BrutalBirdie:
# CLOSED DUE TO INACTIVITY

Please let me know if you want to have this PR/Issue re-opened.

PR #236, opened 2021-08-31
convert unix LF to dos CRLF in logback.xml

crosscodr:
Although I would prefer LF over CRLF, upstream does not so for
logback.xml. After an upgrade of bbb the ansible-role overwrites the
whole file, instead of just changing the loglevels.
It is hard to see the actual diffrences made to the file.
So we should follow upstream here.
This refers to the older PR ebbba-org/ansible-role-bigbluebutton/pull/215.


EdJoPaTo:
Maybe an .editorconfig or some other tooling might be helpful here to ensure the file stays that way?

crosscodr:
> Maybe an .editorconfig or some other tooling might be helpful here to ensure the file stays that way?

Can .editorconfig be used for a single file? I search in the whole repo, files/logback.xml would be the only file with CRLF line terminators.

EdJoPaTo:
It uses globs and globs can also match a single file.
But it might also be a good idea to keep some generic formatting for the yml files for example. Not sure if there is tooling to check for it being correctly used on PRs via GitHub Actions.

The example even matches exact files: https://editorconfig.org/

crosscodr:
Okay, nice idea. I'll add a simple .editorconfig to this PR to start with.

BrutalBirdie (approved):

PR #256, opened 2021-10-25
Fix kurento tasks

### LINKED ISSUE #254: Kurento-Tasks are not idempotent
**Describe the bug**
The tasks in tasks/kurento.yml overwrite themselves each time the playbook is run.

**To Reproduce**
Run this role repeatedly (tag: kurento)

**Expected behavior**
When no changes to kurento configs are made, a following run of ansible should not change anything.

crosscodr:
closes #254 
This is an alternative, cleaner solution to #254, since it uses systemd-templates.
However, it comes with breaking changes. One should probably remove (old) kurento service-definitons before applying this changes.


crosscodr:
I close this PR for now, because the other alternative was merged (#255).
If one day bbb-conf will support templated kurento-services, we can rebase this PR.

PR #258, opened 2021-10-27
Template UFW user rules

### LINKED ISSUE #220: ufw rules are not reset after rerunnning the play

ChrisKar96:
Alternative to #252, without removing UFW
closes #220

### DIFF
--- a/templates/ufw/user6.rules.j2
+++ b/templates/ufw/user6.rules.j2
@@ -1,0 +1,67 @@
+*filter
+:ufw6-user-input - [0:0]
+:ufw6-user-output - [0:0]
+:ufw6-user-forward - [0:0]
+:ufw6-before-logging-input - [0:0]
+:ufw6-before-logging-output - [0:0]
+:ufw6-before-logging-forward - [0:0]
+:ufw6-user-logging-input - [0:0]
+:ufw6-user-logging-output - [0:0]
+:ufw6-user-logging-forward - [0:0]
+:ufw6-after-logging-input - [0:0]
+:ufw6-after-logging-output - [0:0]
+:ufw6-after-logging-forward - [0:0]
+:ufw6-logging-deny - [0:0]
+:ufw6-logging-allow - [0:0]
+:ufw6-user-limit - [0:0]
+:ufw6-user-limit-accept - [0:0]
+### RULES ###
+
+### tuple ### limit tcp {{ bbb_ssh_port }} ::/0 any ::/0 in
+-A ufw6-user-input -p tcp --dport {{ bbb_ssh_port }} -m conntrack --ctstate NEW -m recent --set
+-A ufw6-user-input -p tcp --dport {{ bbb_ssh_port }} -m conntrack --ctstate NEW -m recent --update --seconds 30 --hitcount 6 -j ufw6-user-limit
+-A ufw6-user-input -p tcp --dport {{ bbb_ssh_port }} -j ufw6-user-limit-accept
+
+### tuple ### allow tcp {{ bbb_ssh_port }} ::/0 any ::/0 in
+-A ufw6-user-input -p tcp --dport {{ bbb_ssh_port }} -j ACCEPT
+
+### tuple ### allow tcp 443 ::/0 any ::/0 in
+-A ufw6-user-input -p tcp --dport 443 -j ACCEPT
+
+### tuple ### allow tcp 80 ::/0 any ::/0 in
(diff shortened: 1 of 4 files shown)


templates/ufw/user6.rules.j2
```diff
@@ -0,0 +1,69 @@
...
+
+### tuple ### allow any {{ bbb_coturn_port_tls }} ::/0 any ::/0 in
+-A ufw6-user-input -p tcp --dport {{ bbb_coturn_port_tls }} -j ACCEPT
+-A ufw6-user-input -p udp --dport {{ bbb_coturn_port_tls }} -j ACCEPT
+
+### tuple ### allow udp {{ bbb_coturn_min_port }}:{{ bbb_coturn_max_port }} ::/0 any ::/0 in
+-A ufw6-user-input -p udp -m multiport --dports {{ bbb_coturn_min_port }}:{{ bbb_coturn_max_port }} -j ACCEPT
+
+{% endif %}
+### tuple ### allow udp {{ bbb_freeswitch_min_port }}:{{ bbb_freeswitch_max_port }} ::/0 any ::/0 in
```
crosscodr, line 46: Hi @ChrisKar96, I think you forgot to remove the variable `bbb_freeswitch_min_port` also from the comments :)

ChrisKar96: Nice catch again! Fixed it. I figured that we didn't really need to have new variables for the freeswitch ports, so I opted to revert that change.

ChrisKar96:
After the last force push, UFW is truly idempotent, meaning that not changing any rules results in 'OK' response from ansible, instead of 'changed'.

BrutalBirdie:
I am not good with these firewall templates if someone can verify them I am happy to approve the changes.

crosscodr:
Hi @ChrisKar96, could you please rebase this PR on current bbb/2.3 branch instead of merging it?
In this state of the PR, one can not easily see whats really new in this PR.
Thank you :)

ChrisKar96:
> 
> 
> Hi @ChrisKar96, could you please rebase this PR on current bbb/2.3 branch instead of merging it? In this state of the PR, one can not easily see whats really new in this PR. Thank you :)

Sorry about that, I was tired last night and did a fast forward without realizing. I fixed it now.

BrutalBirdie (approved):
Tested this PR and deployed to https://meet.bigbluebutton.dev/
All working produced an idempotent state for ufw.

+1
LGTM! Nice Job ❤️

ChrisKar96:
As @BrutalBirdie was reviewing this, he discovered a breaking change that I hadn't thought of. So, right now with the PR, any role executed before BigBlueButton which sets ufw rules gets reset since BBB is templating the user.rules directly instead of adding them via the ufw commands. This means that rules applied before or after executing the role will only survive until the next execution of the role. The rules applied after the role break the idempotency of UFW task, while rules applied before are overwritten.

BrutalBirdie (commented):
In discussion because of the breaking change.

BrutalBirdie:
A Role should never have the directive to reset firewall changes done before him, since the Role can't know if the rules are valid or not.
Only the rules provided by the role itself should be added, changed or deleted by the role.
The addition of the UFW to the Role was also just the intention to add some basic security for copy-cats who just use this Role alone and don't know any better.

More so Greenlight and Coturn should be remove into separate Roles since they are not BBB specific. (Yes Greenlight kind off but its not needed for BBB to work.)

The Issue [#220](https://github.com/ebbba-org/ansible-role-bigbluebutton/issues/220) is correct but they way of templating the user.rules file is overreaching its boundaries.

Any thoughts? Please get involved :D

BrutalBirdie:
Sorry, I have struggled with this PR a lot and came to the conclusion that we should not do it this way.

PR #260, opened 2021-10-29
bbb_greenlight_users fix

olwe0002:
somehow the current config with bbb_greenlight_users array set to 
`- { name: '', email: '', password: '', type: 'user' } `
and then use this array with 
`| flatten(levels=1)` in `the Create greenlight users` task leads to the ansible task result "changed":

`TASK [ebbba.bigbluebutton : Create greenlight users] ***************************
changed: [server] => (item={u'password': u'secret', u'type': u'admin', u'name': u'admin', u'email': u'email@company.tld'})`

but the login with the defined credentials is not possible afterwards.

also in the database the user is not contained after this step, checked it with

 `docker exec greenlight-v2 psql "postgresql://postgres:<hereyourpw>@db:5432/<hereyourdbname>" -c "select * from users;" `

but with a "real" array and without flatten filter, it works fine. Maybe the order of the attributes gets mixed by the flatten filter?

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -36,13 +36,13 @@ bbb_turn_servers:
 
 bbb_greenlight_enable: true
 bbb_greenlight_users: []
-  # - name: '' 
-  #   email: '' 
-  #   password: '' 
-  #   type: 'user' 
-  # - name: '' 
-  #   email: '' 
-  #   password: '' 
+  # - name: ''
+  #   email: ''
+  #   password: ''
+  #   type: 'user'
+  # - name: ''
+  #   email: ''
+  #   password: ''
   #   type: 'admin'
 
 bbb_webhooks_enable: false
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -36,8 +36,14 @@ bbb_turn_servers:
 
 bbb_greenlight_enable: true
 bbb_greenlight_users: []
-#  - { name: '', email: '', password: '', type: 'user' }
-#  - { name: '', email: '', password: '', type: 'admin' }
+  # - name: '' 
+  #   email: '' 
+  #   password: '' 
+  #   type: 'user' 
+  # - name: '' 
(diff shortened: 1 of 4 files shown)


crosscodr:
Hi @olwe0002, thank you for your PR.
The flatten filter was used to reflect the behavior of `with_items` when using `loop` instead.
See https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#with-items.
I tested the default vars with `flatten` and without. For me it made no difference:
```
- hosts: localhost
  vars:
    bbb_test:
      - { name: 'user1', email: 'user1@example.com', password: 'user1@example.com', type: 'user' }
  
  tasks:
    - debug:
        msg: "{{ bbb_test | flatten(levels=1) }}"
    - debug:
        msg: "{{ bbb_test }}"
```

@olwe0002 which ansible-version do you use?

BrutalBirdie:
> but the login with the defined credentials is not possible afterwards.
This is a known race condition error.
We have seen this error come and go multiple times now.

Problem is while the container is spinning up ansible tries to create the first user and soft fails with no exit.
Then the second or third user is working.

I know there was a discussion somewhere how to handle this type of race condition but I cant remember where.

crosscodr:
Thank you for clarification @BrutalBirdie. I think this PR would not fix this.
Maybe this can be fixed by using the [pause](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/pause_module.html) module.
@olwe0002 would you mind building and testing a new fix. If so, please rebase this PR and cleanup the commit history. Otherwise, you can open a new PR.

Davka:
In fact, I used the pause module when testing the role with greenlight.

olwe0002:
ok the problem is that rake didn't create all tables yet, I saw this a debug output. I'll close this MR, rebase and create a new MR with a "db:abort_if_pending_migrations" loop. thanks folks

PR #262, opened 2021-11-03
wait until rake did all necessary migrations before greenlight user creation

olwe0002:
the root cause of not created users is, like mentioned [here](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/260), that rake did not create all tables yet when the user is created.

so this MR waits until rake is done, then the user creation works fine.

### DIFF
--- a/tasks/greenlight.yml
+++ b/tasks/greenlight.yml
@@ -37,12 +37,23 @@
         pull: "{{ bbb_greenlight_image_pull }}"
         project_src: /root/greenlight/
 
+    - name: wait until rake did all necessary migrations
+      community.docker.docker_container_exec:
+        container: greenlight-v2
+        command: bundle exec rake db:abort_if_pending_migrations
+      register: greenlight_rake
+      until: greenlight_rake.rc == 0
+      retries: 10
+      when: not ansible_check_mode
+
     # https://docs.bigbluebutton.org/greenlight/gl-admin.html#creating-accounts
     - name: Create greenlight users
-      command: "docker exec greenlight-v2 bundle exec rake user:create[\"{{ item.name }}\",\"{{ item.email }}\",\"{{ item.password }}\",\"{{ item.type }}\"]"
+      community.docker.docker_container_exec:
+        container: greenlight-v2
+        command: "bundle exec rake user:create[\"{{ item.name }}\",\"{{ item.email }}\",\"{{ item.password }}\",\"{{ item.type }}\"]"
       loop: "{{ bbb_greenlight_users | flatten(levels=1) }}"
       register: usercreate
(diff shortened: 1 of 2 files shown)


crosscodr (changes requested):
Hi @olwe0002, thank you for submitting this PR.
I have a suggestion for this task, how about using this module instead of command:
https://docs.ansible.com/ansible/latest/collections/community/docker/docker_container_exec_module.html

olwe0002:
ok, I did that with my new change. I also changed "Create greenlight users" to the same module and improved the changed_when condition. please review

crosscodr:
Nice, looks good to me! Did you test this against a clean setup?

tasks/greenlight.yml
```diff
@@ -37,12 +37,22 @@
...
     - name: Create greenlight users
-      command: "docker exec greenlight-v2 bundle exec rake user:create[\"{{ item.name }}\",\"{{ item.email }}\",\"{{ item.password }}\",\"{{ item.type }}\"]"
+      community.docker.docker_container_exec:
+        container: greenlight-v2
+        command: "bundle exec rake user:create[\"{{ item.name }}\",\"{{ item.email }}\",\"{{ item.password }}\",\"{{ item.type }}\"]"
       loop: "{{ bbb_greenlight_users | flatten(levels=1) }}"
       register: usercreate
       failed_when: "'Invalid Arguments' in usercreate.stdout"
-      changed_when: "'Account with that email already exists' not in usercreate.stdout"
+      changed_when: "'Account successfully created' in usercreate.stdout"
```
crosscodr, line 57: As a normal command, this can not be run in check mode and fails if not explicitly 
```suggestion
      when: not ansible_check_mode

```

crosscodr (changes requested):

tasks/greenlight.yml
```diff
@@ -37,12 +37,22 @@
...
         project_src: /root/greenlight/
 
+    - name: wait until rake did all necessary migrations
+      community.docker.docker_container_exec:
+        container: greenlight-v2
+        command: bundle exec rake db:abort_if_pending_migrations
+      register: greenlight_rake
+      until: greenlight_rake.rc == 0
+      retries: 10
+
```
crosscodr, line 47: As a normal command, this can not be run in check mode and fails if not explicitly 
```suggestion
      when: not ansible_check_mode

```

crosscodr (changes requested):

olwe0002:
yes, i did test it against a clean setup and it works. also I accepted your proposed change and squashed them into one single commit.

crosscodr (approved):

BrutalBirdie (approved):

PR #264, opened 2021-11-08
use bbb_default_logo_url as branding image setting

olwe0002:
This MR sets the bbb_default_logo_url, if the bbb_use_default_logo is set to true, as the branding image, so that on the start page of bbb the logo is visible too.

### DIFF
--- a/tasks/greenlight.yml
+++ b/tasks/greenlight.yml
@@ -56,4 +56,11 @@
       failed_when: "'Invalid Arguments' in usercreate.stdout"
       changed_when: "'Account successfully created' in usercreate.stdout"
 
+    - name: Set greenlight branding image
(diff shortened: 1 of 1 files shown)


olwe0002:
I have to test this first against a clean setup, not sure if the features record is existing at this point, maybe it has to be inserted first. I'll keep you posted

olwe0002:
with the help of a rails expert, I did improve the task. Now a rails command is used to update the value, instead of a direct database update. This works in each case and is much cleaner. please review

olwe0002:
fails on a clean setup, I need to investigate

olwe0002:
now it works also on a clean setup

BrutalBirdie (approved):

Davka (approved):

PR #269, opened 2021-11-21
Juju4 devel 23b - PR #182

BrutalBirdie:
All credits go to @juju4 - please view the discussion for more information.

#182


.github/workflows/default.yml
```diff
@@ -0,0 +1,154 @@
+---
+name: AnsibleCI
```
EdJoPaTo, line 2: Maybe the filename and name should be more descriptive of what the goal of this workflow is. "default" is not saying anything and "AnsibleCI" is not that helpful for a CI workflow in an Ansible role repo. Maybe "Full Deploy" or something like that?

.github/workflows/default.yml
```diff
@@ -0,0 +1,154 @@
...
+          ansible -i inventory --connection=local -m setup localhost
+        if: ${{ always() }}
+        continue-on-error: true
+      - name: After script - firewall
+        run: |
+          sudo iptables -L -vn
+          sudo ip6tables -L -vn
+          sudo ufw status verbose
+        if: ${{ always() }}
+        continue-on-error: true
```
EdJoPaTo, line 98: Personally I like to keep one liners to the top and have the long thingies at the bottom. I think its easier to read that way. (If changing that it should be changed for the whole file, the diff is only an example of what I meant)

```suggestion
      - name: After script - firewall
        if: ${{ always() }}
        continue-on-error: true
        run: |
          sudo iptables -L -vn
          sudo ip6tables -L -vn
          sudo ufw status verbose
```

.github/workflows/default.yml
```diff
@@ -0,0 +1,154 @@
+---
+name: AnsibleCI
+
+on:
+  push:
+  pull_request:
+  # schedule:  # run weekly, every monday 01:00
+  #   - cron: '0 1 * * 1'
```
EdJoPaTo, line 8: A `schedule` will only run when the file is on the default branch. It wont run yet while still a pr / branch.
If it should be running in the future this can be commented in already. If it should not be used it should be removed completely rather than just being a comment.

Also a minor idea: Monday is probably the most crowded day for CI. Maybe do it on Sunday evening?

.github/workflows/default.yml
BrutalBirdie: completely true.
I will change that.

.github/workflows/default.yml
BrutalBirdie: Hmmm I understand the mindset behind that.
I will have a look.

.github/workflows/default.yml
BrutalBirdie: I don't see why running a full deployment on a cron base would make any sense :D
So removing it completely is the best option.

EdJoPaTo: Maybe new releases of BBB packages might introduce changes that might break this role. But as most configs are overwritten anyway I doubt that might have a big impact.

Thats the only thing that came to my mind when thinking about the `schedule` so far.

BrutalBirdie:
@EdJoPaTo If done some changes, could you review them again?

molecule/default/verify.yml
```diff
@@ -28,33 +27,84 @@
...
+    assert:
+      that:
+        - ansible_facts.services[item]['state'] == 'running'
+        - ansible_facts.services[item]['status'] == 'enabled'
+    with_items:
+      - bbb-apps-akka.service
+      - bbb-html5-frontend@1.service
+      - bbb-html5-frontend@2.service
+      - bbb-html5-backend@1.service
+      - bbb-html5-backend@2.service
```
Davka, line 84: That fits, as long as you have 2 front ends and 2 back ends running. The number can, however, be changed.

molecule/default/verify.yml
```diff
@@ -28,33 +27,84 @@
...
+      - bbb-html5.service
+      - bbb-rap-caption-inbox.service
+      - bbb-web.service
+      - bbb-webrtc-sfu.service
+      - coturn.service
+      - kurento-media-server.service
+      - mongod.service
+      - nginx.service
```
Davka, line 92: With regard to the 2.4 one should also check mediasoup

molecule/default/verify.yml
BrutalBirdie: default is 2, and the defaults are getting checked.

molecule/default/verify.yml
BrutalBirdie: lets but this to the backlog when 2.4 is ready

Davka (approved):

PR #270, opened 2021-11-22
[Security] Set Freeswitch default password, Checkvars step, README

BrutalBirdie:


Davka (approved):

Davka (approved):

tasks/checkvars.yml
```diff
@@ -62,6 +62,9 @@
       debug:
         msg: "Please disable monitoring for Systemd or Docker, these can not run alongside"
       failed_when: bbb_monitoring_all_in_one_enable and bbb_monitoring_systemd_enable
+    - name: bbb_freeswitch_default_password
+      debug:
+        msg: "value of bbb_freeswitch_default_password variable is: {{ bbb_freeswitch_default_password variable }}"
```
crosscodr, line 67: Hi @BrutalBirdie, why do we need to print these passwords here, at all?
When I look at the second commit in this PR, I understand, that you just want to fail if the variable is not set.
This can be done otherwise:
```suggestion
    - name: bbb_freeswitch_default_password
      debug:
        msg: "Please set a value for bbb_freeswitch_default_password variable."
      failed_when: bbb_freeswitch_default_password is not defined
      when: bbb_freeswitch_default_password is not defined        
```

crosscodr (changes requested):

crosscodr (approved):

PR #275, opened 2021-12-07
Fix #272 - Purge Greenlight

### LINKED ISSUE #272: Remove Greenlight and move into a separate Role
Greenlight is used for BigBlueButton but does not belong to BigBlueButton by default.
This should be in its own Role.

BrutalBirdie:


crosscodr:
LGTM. Since this is a breaking change, we should make a hint in the README, to look for an extra greenlight role to use.

Davka (approved):

PR #276, opened 2021-12-07
Fix #273 - Purge coturn

### LINKED ISSUE #273: Remove Coturn and move into a separate Role
Coturn is used for BigBlueButton but does not belong to BigBlueButton by default.
This should be in its own Role.

BrutalBirdie:


Davka (approved):

README.md
```diff
@@ -12,7 +12,8 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 
 ---
 
-ℹ️ - PR [#275](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/275) removed the option of installing Greenlight. Please checkout another projoct for Greenlight support
+- ℹ️ PR [#275](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/275) removed the option of installing Greenlight. Please checkout another projoct for Greenlight support
+- ℹ️ PR [#276](https://github.com/ebbba-org/ansible-role-bigbluebutton/pull/276) removed the option of installing coturn. Please checkout another projoct for coturn support
```
EdJoPaTo, line 16: In an ideal situation there should be a "simple does it all" example. With this role it was possible to set up a complete standalone working BBB instance. Throwing out required stuff for a standalone installation will make it harder to set up BBB.
I am not saying this should be a "this role does everything" like it used to and separation of concern is good, but there should be a similar "get a standalone instance with this example" or "Use these 3 thingies and it works".

"checkout another project" without any hint is not very helpful for this.
For greenlight crosscodr suggested his own role in https://github.com/ebbba-org/ansible-role-bigbluebutton/issues/272#issuecomment-984535934 which could be easily linked here.
For coturn there are other roles too (which I never used because this rule did just fine). Something that is known to work should be linked here.
Then its easier to get started.

BrutalBirdie: Fully understandable, I am already working out a solution for this.
But for now I want this PR merged for the next release.

michaelbaisch: Just wanted to ask if there is any update on this? I see there is this role: https://github.com/ebbba-org/ansible-role-coturn with two pull requests that might work as drop in replacement?

PR #278, opened 2021-12-10
add task to replace mongo tmpfs size in startscript

Davka:


README.md
```diff
@@ -53,6 +53,8 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.3/install.html#
 | | `bbb_ice_servers` | a list of RemoteIceCandidate for STUN | `[]` | in array with key `server` |
 | | `bbb_turn_servers` | a list of TURN-Server to use | `{{ bbb_hostname }}` with `{{ bbb_coturn_secret }}` | take a look in defaults/main.yml |
 | | `bbb_greenlight_enable` | enable installation of the greenlight client | `yes` | |
+| | `bbb_mongodb_version` | version of mongodb to be installed | `4.2` | | 
+| | `bbb_mongodb_tmpfs_size' | tmpfs size for the mongodb | `512m` | |
```
BrutalBirdie, line 57: a description when to increase of decrease this value would be usefull

tasks/mongodb.yml
```diff
@@ -22,6 +22,16 @@
...
+- name: set tmpfs size for mongodb
+  become: true
+  lineinfile:
+    path: /usr/share/meteor/bundle/mongod_start_pre.sh
+    regexp: "^  mount -t tmpfs"
+    line: "  mount -t tmpfs -o size={{ bbb_mongodb_tmpfs_size }} tmpfs /mnt/mongo-ramdisk"
+  notify:
+    - stop mongo
+    - restart mongo
+
```
BrutalBirdie, line 34: maybe template this script?
Or is that too much?

crosscodr: I would go for a template, too. Then you can easily see the context, just by looking at the role.
Why do you stop and then restart mongo?

PR #291, opened 2022-02-04
allow to change default bbb_webrtc_sfu_multikurento value without copy&paste

einhirn:
Ansible doesn't allow something like 
```
x=100 #defaults/main.yml

x=x+1 #groupvars/bbb-conf.yml
```
or rather
```
bbb_webrtc_sfu_multikurento: [default value]
bbb_webrtc_sfu_multikurento: "{{ bbb_webrtc_sfu_multikurento | combine({'conference-media-specs':{'OPUS':{'maxaveragebitrate': '64000'}}})  }}"
```
because it references a variable that is being defined at that time. This results in an error `recursive loop detected in template string`.

To make something like this work, I put the variable default value to an underscore-variable first and define the regular variable from that. I.e.
```
_x=100 #defaults/main.yml
x=_x   #defaults/main.yml   

x=_x+1 #groupvars/bbb-conf.yml
```
Now I can use 
```
bbb_webrtc_sfu_multikurento: "{{ _bbb_webrtc_sfu_multikurento | combine({'conference-media-specs':{'OPUS':{'maxaveragebitrate': '64000'}}})  }}"
```
in my group_vars to modify the original defaults without copy&paste...

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,21 @@ bbb_meteor:
         url: "https://{{ bbb_hostname }}/html5log"
 ```
 
+### Modification of `bbb_webrtc_sfu_multikurento`
+
+To add new keys to `bbb_webrtc_sfu_multikurento`, the defaults are actually stored in `bbb_webrtc_sfu_multikurento_default` and then assigned to `bbb_webrtc_sfu_multikurento`.
+
+That way you can add new keys like something like this:
+```yaml
+tempvar:
+  conference-media-specs:
+    OPUS:
+      maxaveragebitrate: "64000"
+
+
+bbb_webrtc_sfu_multikurento: "{{ bbb_webrtc_sfu_multikurento_default | combine(tempvar)  }}"
+```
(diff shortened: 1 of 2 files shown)


einhirn:
I don't know where exactly I'd put a description of this - but I'll try to add something in README.md

BrutalBirdie:
> I don't know where exactly I'd put a description of this - but I'll try to add something in README.md EDIT: I see that `bbb_webrtc_sfu_multikurento` isn't documented at all yet... Should it be? Is it something people need to be made aware of? Or is it something for those of us that pretend to know what they are doing grin ?

True!

Davka (approved):

einhirn:
I've added a paragraph in README and renamed the "internal" variable from `_...` to `..._default` to signify that one is allowed to work with that variable after discussion with @defnull .

BrutalBirdie (approved):

PR #292, opened 2022-02-04
Allow replacing default logo

einhirn:
The role can already set a default logo URL, but it can't yet copy a file to the server to be used as default logo. With this PR it can.


BrutalBirdie (approved):
LGTM!

Davka (approved):

PR #290, opened 2022-01-28
refine logging, remove unused vars and add more bbb-web.properties

Davka:

PR #294, opened 2022-02-07
Combine tasks and fixed conditionals

Nemental:
- combined multiple tasks and added loop
  - to keep task-files short but readable
- fixed conditionals

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -47,8 +47,8 @@
     group: root
     mode: '0600'
   loop:
-    - bbb-apps-akka
-    - bbb-fsesl-akka
+    - bbb-apps-akka.conf
+    - bbb-fsesl-akka.conf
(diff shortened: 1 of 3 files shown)


Nemental:
closes #297

BrutalBirdie:
Wait.. What?
In PR #298 the option `#jinja2: trim_blocks:False` was added and in your PR you delete them again?
I am confused since you stake your PR also `fixes https://github.com/ebbba-org/ansible-role-bigbluebutton/issues/297`

?

Nemental:
@BrutalBirdie My changes will make `#jinja2: trim_blocks:False` obsolete, because we're both aiming the same target. The root cause of the problem is the templating issue merged in #290.

BrutalBirdie (approved):

Davka (approved):

PR #299, opened 2022-02-11
nodejs 12 to 14

Nemental:
- https://docs.bigbluebutton.org/2.4/new.html#whats-new
>Under the hood, BigBlueButton 2.4 installs on Ubuntu 18.04 64-bit, and the following key components have been updated
  Node 14.x (LTS release of Node)


Davka:
We have to follow BBB guidelines and continue to install Node-12 (see BBB-Install)

Node14 is installed with bbb-html5 and is not affected by the setting here.

Issue #304, opened 2022-04-01
Upgrade from 2.3 to 2.4 fails


Davka:
**Describe the bug**
When updating from a 2.3, it can happen that the NGINX cannot start because the load balancer configuration is not yet available at the time.

**To Reproduce**
Upgrade from 2.3 to latest 2.4 using this role

defnull:
Upgrading anything but patch releases is discouraged. There are _lots_ of places where this can go wrong. I'd always install on a fresh Ubuntu if possible.

BrutalBirdie:
I did not have this issue when upgrading aprox 10x 2.3 Servers to 2.4.
Strange.

PR #305, opened 2022-04-01
Ensures loadbalancer.nginx is installed

Davka:
refs #304


defnull:
This makes it more difficult to actually use the loadbalancer hack. How about making this an ansible variable?

defnull:
Or set the variable to a zero-length string in `vhost.conf` just before the include line at the end. This way a custom config file can re-define the nginx variable.

PR #306, opened 2022-04-27
Add option to install even if there are current sessions

Nudin:
Sometimes there are sessions that were not terminated even though they are empty (Probably due to a bug); and sometimes for various reasons it doesn't matter if meetings are stopped. For this there should be a switch to continue even if current meetings are found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.4/install.html#
 | | `bbb_turn_enable` | enable the use uf TURN in general | `yes` | |
 | | `bbb_stun_servers` | a list of STUN-Server to use | `{{ bbb_hostname }}` | an array with key `server` - take a look in defaults/main.yml |
 | | `bbb_ice_servers` | a list of RemoteIceCandidate for STUN | `[]` | in array with key `server` |
-| | `bbb_turn_servers` | a list of TURN-Server to use | `{}` | take a look in defaults/main.yml |
+| | `bbb_force_install` | install even if meetings are running | `no` | current meetings will be terminated |
 | | `bbb_mongodb_version` | version of mongodb to be installed | `4.2` | |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.4/install.html#
 | | `bbb_turn_enable` | enable the use uf TURN in general | `yes` | |
 | | `bbb_stun_servers` | a list of STUN-Server to use | `{{ bbb_hostname }}` | an array with key `server` - take a look in defaults/main.yml |
 | | `bbb_ice_servers` | a list of RemoteIceCandidate for STUN | `[]` | in array with key `server` |
-| | `bbb_turn_servers` | a list of TURN-Server to use | `{}` | take a look in defaults/main.yml |
```
BrutalBirdie, line 57: Why did this doc line get deleted?
Its getting used:
https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/85eb1cce7a95711a49d8489fe6283ad5524106e5/templates/bbb/turn-stun-servers.xml.j2#L43-L50

https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/85eb1cce7a95711a49d8489fe6283ad5524106e5/defaults/main.yml#L18-L28

README.md
```diff
@@ -54,7 +54,7 @@ Also check [Before you install](https://docs.bigbluebutton.org/2.4/install.html#
...
+| | `bbb_force_install` | install even if meetings are running | `no` | current meetings will be terminated |
```
BrutalBirdie, line 57: Just preference discussion.
`bbb_force_install` may read to an unknown user that somehow an installation gets forced.
Wouldn't it make more sense to call this variable `bbb_ignore_running_meetings`?

BrutalBirdie (changes requested):

PR #312, opened 2022-06-14
fix issue #309 when systemd-timesyncd is already disabled

### LINKED ISSUE #309: role will fail when systemd-timesyncd is already disabled
**Describe the bug**
For example, if you use a VM template in which systemd-timesyncd is already disabled or removed, the role will fail

```
Unable to enable service systemd-timesyncd: Failed to enable unit: Unit file /etc/systemd/system/systemd-timesyncd.service is masked.
```

```
"Could not find the requested service systemd-timesyncd
```

**To Reproduce**
Use a VM template where the service is masked or uninstalled

einhirn:
Fixes #309 - or at least it works for me.

https://stackoverflow.com/questions/51765306/using-ansible-to-stop-service-that-might-not-exist/65825886#65825886
shows a way to prevent collecting "services" facts if we don't need to, but I didn't go there yet...

For some reason the service had a state of "not-found" on my target system, that's why there's another check.
```
'systemd-timesyncd.service': {'name': 'systemd-timesyncd.service', 'state': 'stopped', 'status': 'not-found', 'source': 'systemd'},
```

### DIFF
--- a/tasks/ntp.yml
+++ b/tasks/ntp.yml
@@ -9,6 +9,9 @@
   package_facts:
     manager: "apt"
 
+- name: Get the list of services
+  service_facts:
+
 - name: Disable/enable chrony service
   service:
     name: chrony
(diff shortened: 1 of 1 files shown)


defnull:
On a freshly installed `ubuntu:20.04` LXC image `status` is `masked` instead of `not-found`. Could you add that to this fix?

einhirn:
I didn't find a complete list of possible states, so I just added "masked" to a list. Haven't tested it yet, though.

BrutalBirdie (approved):
LGTM!

Davka (approved):

Issue #313, opened 2022-06-14
[BUG] - acme renews certificate on every run


einhirn:
**Describe the bug**
acme.sh renews certificate on every run, which easily results in reaching Let's Encrypt's rate limit.

**To Reproduce**
1. run ansible role at least 5 times on the same host.
2. Error message appears.

**Expected behavior**
If there's already a good certificate, don't force-renew it.

PR #314, opened 2022-06-15
fix issue #313, remove need for acme.sh --force

einhirn:
fix issue #313, remove need for acme.sh --force

"Skip renewal" returns 2, so I accept that as a good return code, too.

PS: Sorry, messed up the commit msg on first try.
PPS: Not properly awake yet, one changed file was still missing in the commit.


BrutalBirdie (approved):
LGTM!

Davka (approved):

PR #317, opened 2022-06-23
Fixed order of imports to prevent fail of creating /etc/bigbluebutton…

GUFFMHRZ:
…/nginx

### DIFF
--- a/tasks/config.yml
+++ b/tasks/config.yml
@@ -24,15 +24,6 @@
     cacheable: true
   when: bbb_secret is not defined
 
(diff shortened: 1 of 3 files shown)


defnull:
Just nitpicking, but I'd move the `Ensure /etc/bigbluebutton/nginx folder exists` task out of config.yaml (where it is not needed) into nginx.yaml instead of swapping their order and keeping the implicit dependency.

BrutalBirdie (approved):
LGTM!

Davka (approved):

PR #319, opened 2022-06-29
Fix ensure freeswitch tls dir exists

Nemental:
First run installation on clean Ubuntu 20.04 throws following error:
```
TASK [ebbba.bigbluebutton : use DTLS-SRTP for freeswitch] ***************************************************************************
fatal: [xx-xx-xx-xx]: FAILED! => {"changed": false, "msg": "Destination directory /opt/freeswitch/etc/freeswitch/tls does not exist"}
```

### DIFF
--- a/tasks/kurento.yml
+++ b/tasks/kurento.yml
@@ -24,6 +24,15 @@
     regexp: 'pemCertificateRSA='
     line: 'pemCertificateRSA=/etc/kurento/dtls-srtp.pem'
 
+- name: Ensure /opt/freeswitch/etc/freeswitch/tls folder exists
+  become: true
+  file:
+    state: directory
+    path: /opt/freeswitch/etc/freeswitch/tls
+    owner: freeswitch
+    group: daemon
+    mode: 0700
+
 - name: use DTLS-SRTP for freeswitch
   become: true
   copy:


defnull:
Interesting, I do not see this problem on a fresh install. The task is part of `kurento.yml` which is triggered after `freeswitch.yml`. I see no way that this directory could be missing unless there are errors before that.

Nemental:
To make sure I'll reset my VM and try it again :)

Nemental:
Please don't ask me why, but it work's... Already tried many times, but maybe I did something weird before.

denzs:
Addition: creating the tls folder wasnt sufficient.. after the folder was created manually the wss.pem was still missing...
can somebody tell me whci script/mechanism should generate that file?

einhirn:
As I wrote before, it's FreeSWITCH that creates the directory and wss.pem. Check your syslog on the failed machines for "apparmor" entries - there should be a message that "freeswitch" wasn't allowed to write to `/opt/freeswitch/etc/freeswitch`. I still don't know why it sometime works and sometime doesn't... But maybe new apparmor rules aren't applied immediately and we encounter some kind of a race condition...

PR #322, opened 2022-07-14
Restore bbb_nginx_listen_https option

znerol:
First introduced in #88 / 2c4385b4 but somehow this option got lost along the way.

### DIFF
--- a/templates/nginx/vhost.conf
+++ b/templates/nginx/vhost.conf
@@ -2,6 +2,8 @@ server {
     listen 80;
     listen [::]:80;
     server_name {{ bbb_hostname }};
+
+{% if bbb_nginx_listen_https %}
     return 301 https://$host$request_uri;
 }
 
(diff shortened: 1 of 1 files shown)


defnull:
While this probably works, it is a very strange or hard-to-read way to modify the nginx config blocks. Could this be refactored to be more clear?

BrutalBirdie:
Ahh, thanks for reporting.
I believe it would be even better to make all template files user overwrite-able.

Also what @defnull said, a refactoring would be nice.

defnull:
Can we just drop HTTP (without S) support? It's 2022 and certificates are free. I'm not sure if BBB even works properly over HTTP as many additional restrictions for XHR or cookies apply. BBB lists SSL under [minimum server requirements](https://docs.bigbluebutton.org/2.5/install.html#minimum-server-requirements)

mmv-ru:
@defnull
 I got your reasons and decide to reconfigure reverseproxy to allow passtrough /.well-known/acme-challenge/ to application host. It allow get letsencrypt work in parallel on reverseproxy and on application host.
But my setup is reverseproxy behimd nat, which share single public IP between multiple domains (applications). :-)

PR #325, opened 2022-09-02
Add option to install even if there are current sessions

Nudin:
Sometimes there are sessions that were not terminated even though they are empty (Probably due to a bug); and sometimes for various reasons it doesn't matter if meetings are stopped. For this there should be a switch to continue even if current meetings are found.

This is an improved version of #306.


BrutalBirdie (approved):
LGTM!
Now just wait for the Github pipeline checks.

BrutalBirdie:
@Nudin can you rebase your commit against the latest changes please?

Davka (approved):

Issue #318, opened 2022-06-27
[BUG] - use disabledFeatures=learningDashboard instead of learningDashboardEnabled=false


defnull:
BBB 2.5 complains about:
`WARN  o.b.api.ParamsProcessorUtil - [DEPRECATION] use disabledFeatures=learningDashboard instead of learningDashboardEnabled=false`

einhirn:
This also applies to "breakoutRoomsEnabled=true".

einhirn:
From discussing this in the matrix channel, it would seem best to either create two dictionaries (`default` and `custom` config) that can be traversed in the template or define a name prefix like `bbb_conf_*` and use 
```lookup('ansible.builtin.varnames', '^bbb_conf_.+')```
 to access them in the template. The latter option would make it easier to override defaults.

PR #335, opened 2023-01-09
[2.5/2.6] Quick workaround for disabling learning dashboard

einhirn:
It only fixes `bbb_learning_dashboard_enabled` for #318 - other features that should use the new syntax for disabling won't be factored in...


timm2k:
Since BBB 2.5.11 there is no need for this workaround.

einhirn:
> Since BBB 2.5.11 there is no need for this workaround.

But the old syntax is still deprecated and will be removed some time later. Now this PR is less of a hack so we'd be ready for it.

einhirn:
Closing this in favor of #357 - it allows all currently available and future "disableFeature"-Options, albeit without syntax check inside the role...

PR #344, opened 2023-03-26
bbb-webrtc-sfu.service: do not set CPUSchedulingPolicy in a container

davidmehren:
BBB 2.6 does not work correctly in LXC containers, as the unit file of `bbb-webrtc-sfu` sets `CPUSchedulingPolicy`, which is not supported there.

This adjusts the already existing override file to unset the option when `bbb_container_compat` is enabled


defnull:
Not really a show-stopper for this RP, but we run BBB on privileged LXC containers that do support setting `CPUSchedulingPolicy` just fine, but still have to disable `bbb_container_compat` because installing `auditd` still fails on those. We then re-enable features disabled by `bbb_container_compat=true` in a later ansible task, and would also have to revert this change for the same reason.

basisbit (approved):

Davka (approved):

PR #349, opened 2023-04-05
Add cluster proxy configuration

tibroc:
This commit adds the capability to configure bbb for the cluster proxy setup.

It is based on the work of @defnull (https://gist.github.com/defnull/998f9217cb56edcc397ec6a1d46bb1c0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,27 @@ bbb_meteor:
           viewer: 8
 ```
 
+### Cluster proxy
+
+This role is capable of configuring BigBlueButton for the (cluster proxy setup)[https://docs.bigbluebutton.org/administration/cluster-proxy/].
+To use this feature you need to define both `bbb_proxy_host` to the url of your cluster proxy and
+`bbb_proxy_name` to the name your BBB host should have in the proxy
+(you can set this e.g. to the subdomain of your BigBlueButton host, as shown in the defaults).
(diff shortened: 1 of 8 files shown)


defnull:
Thanks for FINALLY getting this into the role. I wanted to donate our code for ages now, but never found the time to write a proper PR. Some feedback:

* Some tasks are missing `when: bbb_proxy_url is defined` I think? Role might break if proxy setup is not enabled.
* Just nitpicking, but I'd choose different config variable names:
    * `bbb_proxy_host` (instead of `bbb_proxy_url`) because it is just a hostname, not a full URL.
    * `bbb_proxy_name` (instead of `bbb_proxy_lb_name`) because it is a name for the BBB node, not for the loadbalancer.
* I'd remove the default for `bbb_proxy_name`. Using the subdomain is clever (and we do that for our setup) bit not guaranteed to work for everyone.

BrutalBirdie:
hey @tibroc could you please update the PR to the latest state? <3

tibroc:
@BrutalBirdie: Done. However, it might be worth noting that I did not test it against any other version than 2.6.

BrutalBirdie:
That is totally fine.
The current master / 2.6 tag is als not compatible to deploy a 2.5 BBB.
For this the 2.5 branch / tags should be used.

PR #353, opened 2023-04-24
Feature/dialin scalelite

Nemental:
https://github.com/blindsidenetworks/scalelite/blob/master/dialin-README.md

https://developer.signalwire.com/freeswitch/FreeSWITCH-Explained/Modules/mod_xml_curl_1049001/#installation


simoncolincap:
We implemented this for our BBB servers by disabling dial-in in this role and then adding two tasks outside of it to copy the dial-plan and allow the SIP connection in the firewall. The rest of the configuration was done in Freeswitch and Scalelite.
@Nemental Let me know if you have any questions about how we did it

PR #354, opened 2023-05-10
Added $scheme-Variable to nginx-configuration.

GUFFMHRZ:
Pull request to reflect changes from 
https://github.com/bigbluebutton/bigbluebutton/pull/17732
https://github.com/bigbluebutton/bbb-install/pull/642


ldroege (approved):
I'd prefer #355, but this already helps if nginx itself manages SSL connections.

PR #357, opened 2023-05-24
Implement disabledFeatures and remove old flags

### LINKED ISSUE #318: [BUG] - use disabledFeatures=learningDashboard instead of learningDashboardEnabled=false

tibroc:
This commit implements the disabledFeatures List and removes some old options. The deprecated way of disabling the learning dashboard and breakout rooms will still work in the role. But the resulting bbb-web.properties file will already comply with the new method.

This solves at least a part of #335 

Closes #318


einhirn:
Perhaps we could add a list of allowed features and check whether `bbb_disabled_features` contains only those on that list? That way the deployment would fail in Ansible and not in BBB in case of e.g. a typo.

defnull:
Nitpicking, but `whatever is false` could be written as just `whatever` in most places.

BrutalBirdie (approved):

einhirn:
Hmm... Sorry for re-animating this thread - I just tried to only disable "LearningDashboard" but it also disabled "breakoutrooms"... Have to investigate - anyone else with this issue?

einhirn:
This line has inverted logic:
https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/299fb20a24506cd3f69329f862bb6f6022076a55/defaults/main.yml#L190
It puts "breakoutrooms" into disabled features when the deprecated "enable breakoutrooms"-Flag is set to true.
Same with "learning dashboard"...

PR #366, opened 2023-09-13
Only run certificate related tasks when https is enabled

louim:
If Nginx will not be listening on port 443, then there is no need to generate certificates. Without this setting, the playbook will fail at the "Ensure bbb_ssl_cert exists" task because no SSL certificate will have been generated.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -48,6 +48,7 @@
   tags: mongodb
 
 - import_tasks: certificate.yml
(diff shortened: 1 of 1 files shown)


BrutalBirdie:
Nice catch! Thanks for your contribution <3

BrutalBirdie (approved):

Davka (approved):

PR #368, opened 2023-11-19
Fix 'dubious ownership' when installing prometheus exporter

moonlies:
When installing the prometheus monitoring exporter, the role fails when running the playbook for the second time, because the bbb-exporter's home directory, the cloned repository, is owned by the bbb-exporter user, who could modify the contents which could lead to privilege escalation. So ansible stops the playbook at that point. Retaining ownership with the root user keeps bbb-exporter from potentially being able to modify anything and the role finishes successfully.

This problem occurs when running with bbb_monitoring_external: true

### DIFF
--- a/tasks/monitoring/systemd.yml
+++ b/tasks/monitoring/systemd.yml
@@ -15,14 +15,6 @@
   notify:
     - Restart monitoring service
 
-- name: Change ownership directory
-  file:
-    path: "{{ bbb_monitoring_systemd_directory }}"
-    state: directory
-    owner: bbb-exporter
-    group: bbb-exporter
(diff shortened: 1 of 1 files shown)


BrutalBirdie:
Odd, I always use `bbb_monitoring_external` but it never crashed on my on a re-run.
I will check this out.

BrutalBirdie:
Now I know why this never occurred to me.
I always use the `bbb_monitoring_all_in_one_directory` option which then installs it in `/root/bbb-monitoring` and runs in docker.

Will test again.

BrutalBirdie (approved):

PR #380, opened 2025-02-03
Major rewrite for BBB 3.0

defnull:
This is a WIP to get this role in shape for BBB 3.0 (and future versions).

Please note:
* <del>This PR targets `bbb/2.7` because there is no appropriate target branch in the ebbba-org repository yet and it is based on `bbb/2.7`. **No not merge!**</del>
* The role will change in backwards incompatible ways, which should be fine for a major release upgrade.
* I will use this opportunity to clean up, simplify, modernize and refactor this role in major ways and I will not try to keep this PR small or easy to review. This will be a major rewrite, not a small patch. If you think that this should be a fork and not a PR, please tell me.
* If you want to contribute to this PR, please do so by forking my fork and sending PRs or patches my way for now. <del>There is no appropriate branch or PR target in the ebbba-org repository yet.</del>
* Goals: Port to BBB 3.0. clean up role to simplify future development and maintenance. Add missing core features.
* Non-goals: Support every use case ever. Be backwards compatible to 2.7. Allow major in-place BBB upgrades. Support multiple BBB releases from a single branch.

## Progress

- [x] Revise and fix all tasks until BBB 3.x installs successfully. Simplify and refactor along the way.
- [x] Implement missing core features.
  - [x] Cluster setup.
  - [x] Embedded coturn.
  - [x] Manage optional recording formats (minimum: video).
  - [x] Better support privileged containers (LXC).
- [ ] Whishlist: Nice-to-have features.
  - [x] Custom webcam backgrounds.
  - [x] Better fonts in libreoffice container.
  - [x] Fixed user and group IDs.
  - [x] Install various tools (e.g. bbbctl).
  - [ ] Support clamav and presentation converter hooks.
  - [x] Add `fix.yml` tasks for various BBB bugs/workarounds.
- [ ] Rework test framework and tests.
- [ ] Rework documentation.
  - [x] Replace the unreadable config parameter table with something else.
  - [x] Add "getting Started" section with a minimal example.
  - [ ] Add examples for more involved topics (e.g. cluster setup).
- [x] Soft targets
  - [x] Better sanity checks and safeguards.
  - [x] Improve role performance, reduce unneeded changes or handler invocations.
  - [x] Role should be idempotent. It should report zero changes if run twice.


defnull:
This is now in a usable state. It's still far from feature complete or bug free, but it's a start and it installs BBB most of the time.

defnull:
Embedded coturn works, but I only tested the scenario where the public IP is actually bound to the default interface. The role still needs testing in a NAT environment where `bbb_public_ipv4 != bbb_bind_ipv4`.

BrutalBirdie:
Starting the review today.

defnull:
I'm happy with the current shape and stop tinkering while you have a look, I can continue adding nice-to-have features once it's merged :)

defnull:
It's a dirty hack, but it works surprisingly well! Not sure who to credit for the idea, I took inspiration from the `bbb-install.sh` script and tweaked it a bit (e.g. `tcp-proxy` mode). I think it works well enough to be enabled by default if no external `bbb_turn_servers` are configured.

defnull:
Stuff that I did not properly test yet:

* Servers with no IPv6 Address
* Servers behind NAT where `bbb_bind_ipv4 != bbb_public_ipv4`
* Connecting to a dual-stack server with a client that just speaks IPv6

All this *should* work, but I changed a lot in that regard and do not have a suitable test setup at hand.

BrutalBirdie:
> Connecting to a dual-stack server with a client that just speaks IPv6

This I can test this. I only have IPv4 tunnled and can disable IPv4 in my Router for testing.

defnull:
I'm pretty sure that IPv4-only works and that IPv6-only is not supported by upstream either. NAT would be interesting to test, though. I can fire up an OpenStack VM for that and work on it a bit. But how about merging this anyway (to both bbb/3.0 and master) so this grains some visibility?

BrutalBirdie (approved):
After multiple deployments with migrated settings.
Does not look good to me.

Looks even better to me!
👍🏻

PR #383, opened 2025-05-07
Fix regex for mounting fonts in notes conversion

tibroc:


tasks/mixin-fonts.yml
```diff
@@ -10,7 +10,7 @@
   become: true
   replace:
     path: "{{ item }}"
-    regexp: " -v /usr/share/fonts/(\\\\?):/usr/share/fonts/\\\\?:ro.*? --rm"
```
defnull, line 13: I think I'd prefer a "positive look behind" expression to make sure there is a space in front of the "-v" parameter, but not match (and replace) it: "(?<=\\s)-v..."

defnull (changes requested):
I think I'd prefer a "positive look behind" expression to make sure there is a space in front of the "-v" parameter, but not match (and replace) it:

tasks/mixin-fonts.yml
tibroc: Done in [93a3d81](https://github.com/ebbba-org/ansible-role-bigbluebutton/commit/93a3d81b88a5bd0d15f616a5bff796dcbb994ee8).

defnull (approved):

PR #385, opened 2025-05-08
offer experimental livekit integration

tibroc:
This pull request introduces the option to enable livekit as media streaming backend as currently described in https://docs.bigbluebutton.org/new-features/#integration-with-livekit.

I tried to follow the patterns and structure this ansible role already uses. Please inform me where I might have it misunderstood.

It was a bit more tricky than I initially thought, because livekit-server currently generates an api-key-pair during install (see https://github.com/bigbluebutton/bigbluebutton/blob/v3.0.x-release/build/packages-template/bbb-livekit/after-install.sh#L4) that needs to be taken into account when handling the configuration files in `/etc/bigbluebutton/`.


defnull:
LiveKit API secrets are just random strings, we can generate those ourselves (e.g. using `{{ ('livekit' + bbb_host_seed) | hash('sha256') }}`) and just replace the ones generated by [livekit-server generate-keys](https://github.com/livekit/livekit/blob/aee34ffefad89f135378bf11b8461e866d83e01b/cmd/server/commands.go#L39) with role-controlled values. The key name can also be any random string, as long as it starts with `API`.

If we control the key and secret, we can generate the full config file without slurping and follow the same concept as for the other config files -> merge a role-defined `bbb_config_livekit_base` with a user-defined `bbb_config_livekit` (which can be undocumented for now) and completely overwrite the `/etc/bigbluebutton/livekit.yaml` file.

Would you like to try that? Just make sure to to the other places where the key needs to be replaced are also all managed or fixed by the role (`/etc/default/bbb-webrtc-recorder` and `/etc/bigbluebutton/bbb-webrtc-sfu/production.yml`)

defnull:
Oh and we may need systemd overrides for `livekit-server` and `livekit-sip` to disable `CPUSchedulingPolicy` if `bbb_container_compat` is true. See https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/master/templates/systemd-overrides/bbb-webrtc-sfu.j2 for example

tibroc:
@defnull I took both comments into consideration in 8e732bb and also added variable checks similar to the existing ones.

PR #386, opened 2025-05-23
Feature: Experimental livekit

defnull:
Based on #385 but with additional fixes and adjustments needed after BBB 3.0.7 was releases.


tibroc:
This seems to run correctly on our infra, I can not test sip though at the moment.

defnull:
SIP/Dialin does not work, and I cannot get it to work. But that's fine, it's not marked as ready by upstream anyway.

defnull:
Sorry for the delay, I was sitting on that last commit for a while, frustrated that I could not get dial-in to actually work. But the changes are still useful once upstream is ready for it. Merge to main?

PR #388, opened 2025-08-05
Support custom ownership of SSL files

simoncolincap:
Adds the options to specify ownership of the SSL certificates, this is needed to allow non-root users to handle certificates during renewal.


BrutalBirdie (approved):
LGTM!

defnull (approved):

Issue #389, opened 2025-09-17
[BUG] - Non boolean conditions are deprecated


znerol:
**Describe the bug**

Non boolean conditions are deprecated in ansible and are scheduled to be removed from ansible-core version 2.23.

**To Reproduce**

1. Use a recent enough ansible version (mine is `12.0.0~a6+dfsg-1` from debian trixie)

```
[WARNING]: Deprecation warnings can be disabled by setting `deprecation_warnings=False` in ansible.cfg.
[DEPRECATION WARNING]: Conditional result was '[...]' of type 'str', which evaluates to True. Conditionals must have a boolean result. This feature will be removed from ansible-core version 2.23.
Origin: ebbba.bigbluebutton/tasks/check-vars.yml:47:7

45 - name: Check IPv4 variables
46   when:
47     - bbb_bind_ip4 or bbb_public_ip4
         ^ column 7

Broken conditionals are currently allowed because the `ALLOW_BROKEN_CONDITIONALS` configuration option is enabled.
```

```
[DEPRECATION WARNING]: Conditional result was '[...]' of type 'str', which evaluates to True. Conditionals must have a boolean result. This feature will be removed from ansible-core version 2.23.
Origin: ebbba.bigbluebutton/tasks/check-vars.yml:59:7

57 - name: Check IPv6 variables
58   when:
59     - bbb_bind_ip6 or bbb_public_ip6
         ^ column 7

Broken conditionals are currently allowed because the `ALLOW_BROKEN_CONDITIONALS` configuration option is enabled.
```

```
[DEPRECATION WARNING]: Conditional result was '[...]' of type 'str', which evaluates to True. Conditionals must have a boolean result. This feature will be removed from ansible-core version 2.23.
Origin: ebbba.bigbluebutton/tasks/check-running.yml:17:7

15   when:
16     - bbb_check_running
17     - bbb_old_secret
         ^ column 7

Broken conditionals are currently allowed because the `ALLOW_BROKEN_CONDITIONALS` configuration option is enabled.
```

```
[DEPRECATION WARNING]: Conditional result was '[...]' of type 'str', which evaluates to True. Conditionals must have a boolean result. This feature will be removed from ansible-core version 2.23.
Origin: ebbba.bigbluebutton/tasks/check-running.yml:28:7

26   when:
27     - bbb_check_running
28     - bbb_old_secret
         ^ column 7

[...]

PR #391, opened 2025-09-22
Fixup #389 ansible 12 implicit boolean conversion

BrutalBirdie:


defnull:
Some variables (e.g. `bbb_bind_ip6`) are always defined, but can be `False` or `None` or empty. You can use the custom filter `present` which returns true if the value is defined and a non-empty string or collection (similar to default python behavior).  See https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/master/test_plugins/helper.py#L5

PR #394, opened 2025-11-18
bugfix/dialin_provider_ip_check

wilkis3:
The [documentation](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/master/README.md) said that the variable `bbb_dialin_provider_ip` is only required if  if `bbb_dialin_enable` **and** `bbb_firewall_enable` are true.
Independent that `bbb_firewall_enable` is not defined in any other parts of the role and the default of `bbb_ufw_enable` is false, is this bugfix relevant

### DIFF
--- a/tasks/check-vars.yml
+++ b/tasks/check-vars.yml
@@ -142,7 +142,16 @@
     that: "{{ item }}"
   loop:
     - bbb_dialin_provider is ip or bbb_dialin_provider is fqdn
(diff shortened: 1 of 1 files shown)


BrutalBirdie (approved):
LGTM!

defnull:
Turns out, this was broken because there is no `bbb_firewall_enable`. Documentation was wrong. I'll push a fix that checks for `bbb_ufw_enable` instead.

PR #396, opened 2026-01-07
bugfix/remove-hostname-check

wilkis3:
The README documentation says that `bbb_hostname` has `ansible_fqdn` as default. Why are there any checks? Maybe someone will run a BBB server with an other FQDN than BBB hostname.

### DIFF
--- a/tasks/check-vars.yml
+++ b/tasks/check-vars.yml
@@ -30,7 +30,6 @@
     that: "{{ item }}"
   loop:
     - bbb_secret is defined
-    - bbb_hostname is fqdn
     - bbb_apt_mirror is url
     - bbb_apt_key is url
     - bbb_disabled_features is subset(bbb_allowed_disabled_features)


defnull:
The `bbb_hostname` variable defines the public domain of your BBB host. Public domains are always fully qualified, or clients would not be be able to resolve them. If your BBB host is only reachable in a local network, it should still have a fully qualified domain (e.g. bbb.mycompany.internal). Is there a reason why you want to set it to something that is not an fqdn?

defnull:
Maybe there is a misunderstanding. The check only makes sure that `bbb_hostname` is a fully qualified domain name, it can still be different from `ansible_fqdn` if you change the default.

wilkis3:
> Maybe there is a misunderstanding. The check only makes sure that `bbb_hostname` is a fully qualified domain name, it can still be different from `ansible_fqdn` if you change the default.

Ah, okay. Than is this PR obsolete.

PR #399, opened 2026-05-08
Make blocknote configurable as shared notes editor.

tibroc:
As of BBB v3.0.25 blocknote can be used as a replacement for etherpad in the shared notes. This commit adds the option to select the shared notes editor.

I did not add the option for a config override file for blocknote (compare https://github.com/bigbluebutton/bigbluebutton/pull/24519/changes#diff-3cb944e426b15411f0accb25bd2df64a2901926872c3270894e0d338143b890e), because I do not see the need at the moment.


BrutalBirdie (approved):
LGTM!

defnull:
Nitpick, but can you make `blockNote` lowercase (or change tests to be case-insensitive) and add a test that warns if there is an invalid value to prevent typos there?

<del>Also, if I remember correctly, the proxy mixin would need some additions for this to work in a proxy cluster setup.</del> Maybe move the cluster-proxy stuff to the proxy mixin instead of a separate file? I'd like to keep everything that needs to change for a proxy mode in one file, do be able to compare it with the official documentation more easily.

tibroc:
@defnull I accounted for both comments. What I did not do is make the whole etherpad config optional depending on if you use blocknote. Since I am guessing that blockenote will become the default sooner than later, I am guessing this is ok?

defnull:
Hmmmm. I just noticed that both can be installed at the same time and also used at the same time (dynamically selected during the meeting create call), which means that we actually *have* to keep etherpad installed and configured. So, the PR can be merged I think. I'll try to test and merge this week.

Once blocknote is stable, we can switch things around and make etherpad optional, or remove it entirely. We can be a little bit opinionated with this role I think. If something is clearly the better better, then there is no reason to continue supporting the other options.

PR #400, opened 2026-05-12
add bbb-playback-notes to main.yml

michelleDeko:
I just tested this ansible role and I noticed that the notes format we usually use is missing here. It does add the notes format to `/etc/bigbluebutton/recording/recording.yml` but never installs the bbb-playback-notes package.

This PR fixes this.


defnull (approved):
LGTM

PR #403, opened 2026-06-02
fix: Proxy setup CORS headers

defnull:
BBB 3.0.28 introduced a new nginx config variable named `$bbb_cors_origin` which is used throughout BBB nginx includes as the value for `Access-Control-Allow-Origin` headers and defaults to `*`. Wildcard origins do not allow CORS requests with credentials (e.g. cookies), so this has to be defined for clusters with a proxy.


defnull:
There are more issues, WIP

defnull:
No, the issues were not related to this specific fix. It works as expected.

defnull:
Already merged

PR #404, opened 2026-06-02
ImageMagick policy override for thumbnail generation

einhirn:
[ImageMagick policy override for thumbnail generation](https://github.com/ebbba-org/ansible-role-bigbluebutton/commit/16a84b57c86c91f1481cd3cf7d1174b96917004c)

BBB ships a restrictive policy file for ImageMagick that prevents the re-generation of thumbnails. This adds an override that gets applied only when ImageMagick binaries are executed in the `virtual-backgrounds`-base path.

This file and path could possibly be removed after our work is done, but otoh if anyone can write files somewhere, they still could write their own override file.

### DIFF
--- a/files/ImageMagick-policy.xml
+++ b/files/ImageMagick-policy.xml
@@ -84,6 +84,8 @@
   <!-- don't read sensitive paths. -->
   <policy domain="path" rights="none" pattern="/*"/>
   <!-- allow access to required paths. -->
+  <policy domain="path" rights="read" pattern="/usr/share/bigbluebutton/html5-client/resources/images/virtual-backgrounds/*"/>
+  <policy domain="path" rights="write" pattern="/usr/share/bigbluebutton/html5-client/resources/images/virtual-backgrounds/thumbnails/*"/>
   <policy domain="path" rights="read|write" pattern="/var/bigbluebutton/*"/>
   <policy domain="path" rights="read|write" pattern="/tmp/*"/>
   <!-- Indirect reads are not permitted. -->


einhirn:
This issue only appears during second to Nth runs of the role, because the hardened policy file gets installed after the thumbnails are generated during first run.

defnull:
The ImageMagick policy is actually shipped by [this role](https://github.com/ebbba-org/ansible-role-bigbluebutton/blob/master/files/ImageMagick-policy.xml) because it is usually written by the [install.sh](https://github.com/bigbluebutton/bbb-install/blob/v3.0.x-release/bbb-install.sh#L414) script which this role does not use. Could we just add the necessary rules to the existing policy file?

einhirn:
Then the additional path would be allowed in BBB apps, too. I don't know the exact rationale behind the restrictive policy. If adding the path doesn't circumvent it's desired effect, we could do that, yes.

einhirn:
Hmm... For some reason, the override approach doesn't work on the next system I tried it on. 
**EDIT:** it was an Ubuntu Pro enabled system that used an ESM ImageMagick package build which possibly is restricted even further than the regular version.

Changing the `/etc/ImageMagick-6/policy.xml` file works fine, so I changed this PR to that. If a more restrictive policy is already in place, the role will still fail, but in that case the server should either be re-deployed or have the policy file amended via other means before using this role.

defnull:
I cannot test this (because this seems to only be an issue on Ubuntu Pro with updated packages) but I can also not see any drawbacks of this change now, so: LGTM

defnull (approved):

